PR #879, opened 2022-12-21
Instructions and targets for debugging Qt and PySide6

keithel-qt:
* Add cargo make targets that build and install PySide6 from source
  - prep-debug-pyside
  - build-pyside
  - build-dist-install-pyside
  - build-install-pyside
* Add new README.debug.md describing how to use the debug targets.
* Add some scripts that aid debugging intermittent crashes.
* Remove PySide6 as dependency from pyproject.toml (that flit reads). This is because we now manually install it, as we may install from built wheel, or from pip.


Makefile.toml
```diff
@@ -338,7 +407,14 @@ cp . ../../../py39
...
 exec --fail-on-error ${PYTHON} ./get-pip.py
 exec --fail-on-error ${PYTHON} -m pip install wheel flit . ".[test]" ".[ssh-tunnel]"
-cm_run_task generate-resources
+
+# If pyside-setup exists, then assume user wants to build and install pyside6
+if is_path_exists "${WORKSPACE}/pyside-setup/setup.py"
+  cm_run_task get-qt-version
+  cm_run_task build-install-pyside
+else
+  exec --fail-on-error ${PYTHON} -m pip install PySide6==6.3.2
```
john-michaelburke, line 416: Hardcoding the version here will diverge from our pyproject.toml

keithel-qt: Maybe there is a way to conditionalize the installation of this in flit? so it doesn't get installed if we're doing a debug build, but does otherwise? Not sure how that could be conditionalized.

john-michaelburke: Yeah, we could work something out. Maybe the @swift-nav/devinfra team can help take it over. We would certainly want it to be in lockstep with CI first and foremost.

silverjam: I think a global variable is a good first step, so we don't have to update the version in multiple places.

Makefile.toml
```diff
@@ -338,7 +406,14 @@ cp . ../../../py39
 cd ../../../
 exec --fail-on-error ${PYTHON} ./get-pip.py
 exec --fail-on-error ${PYTHON} -m pip install wheel flit . ".[test]" ".[ssh-tunnel]"
-cm_run_task generate-resources
```
adrian-kong, line 341: `check-all` seems to be missing `generate-resources` files

adrian-kong (approved):

utils/pyside-setup-6.3-includepy-fallback.patch
```diff
@@ -0,0 +1,92 @@
+From bd7b811ed1a5c53b493016847165844ce595f53e Mon Sep 17 00:00:00 2001
+From: Keith Kyzivat <keith.kyzivat@qt.io>
+Date: Wed, 21 Dec 2022 09:07:34 -0500
+Subject: [PATCH] Properly fall back if INCLUDEPY config_var !exist
```
adrian-kong, line 4: will this patch be upstreamed to pyside-setup?

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3889080232)

silverjam (approved):

silverjam:
@adrian-kong please merge at your leisure

PR #862, opened 2022-12-13
event based recording status [CPP-843]

adrian-kong:
removes logging thread, send smaller packets of changes,
defers the full state to the frontend.
we can also refactor this into the event receivers that was introduced in instant filter updates etc... 

downside is we send whenever we receive a frame meaning it would be dependent on the stream of info - which we can do some batching if this happen (too many updates for frontend gui => laggy ?)

this is also to prepare for another ticket - record empty streams of data

PR #885, opened 2022-12-28
path check before log [CPP-851]

adrian-kong:
check before writing to file

- refactors sbp logging struct to manually convert into json bytes to handle the counter changes.
- changes from compactformatter to haskellishfloatformatter
- simple path check (assuming file doesnt get changed in that interval) to stop byte counter


console_backend/src/tabs/main_tab.rs
```diff
@@ -150,7 +139,12 @@ impl MainTab {
         start_recording(&self.client_sender, filepath.display().to_string());
     }
 
-    pub fn serialize_frame(&mut self, frame: &Frame) {
+    /// Serialize frame incoming data,
+    ///
+    /// # Parameters:
+    /// - `frame`: The raw incoming data frame
+    /// - `msg`: Parsed message if present
+    pub fn serialize(&mut self, frame: &Frame, msg: Option<&Sbp>) {
```
adrian-kong, line 147: probably better to just have one or the other parameter but since we want to keep a copy of the computed frame -> msg to prevent recalculation, just keep both params

console_backend/src/output.rs
```diff
@@ -50,42 +50,59 @@ impl CsvLogging {
...
+            SbpLogging::SBP => Some(frame.as_bytes().to_owned()),
+        };
 
-    pub fn new_sbp_json<P: AsRef<Path>>(filepath: P) -> Result<SbpLogger> {
-        Ok(SbpLogger::Json(JsonEncoder::new(
-            File::create(filepath)?,
-            CompactFormatter,
-        )))
-    }
+        if let Some(path) = &self.path {
```
adrian-kong, line 83: not sure how costly this is checking everytime we serialize.
looks like TOCTOU 👀 but it should be ok

or alternatively use some sort of file notify.

console_backend/src/output.rs
```diff
@@ -50,42 +50,59 @@ impl CsvLogging {
...
+}
 
-    pub fn serialize(&mut self, frame: &Frame) -> Result<()> {
-        match self {
-            SbpLogger::Sbp(writer) => writer.write_all(frame.as_bytes())?,
-            SbpLogger::Json(logger) => logger.send(&frame.to_sbp()?)?,
-        }
-        Ok(())
+impl SbpLogging {
+    pub fn new_logger(&self, path: PathBuf) -> Result<SbpLogger<File>> {
```
adrian-kong, line 103: only Files for now, but `W: impl Write` 
🤔 will we ever want to write elsewhere like uploading this sbp on the fly

console_backend/src/tabs/main_tab.rs
adrian-kong: also NOTE: this isn't just serialize, it also updates the sbp formats which might be nice to seperate into a different method and have this only do serialization stuff

console_backend/src/output.rs
```diff
@@ -50,42 +50,59 @@ impl CsvLogging {
...
+    path: Option<PathBuf>,
 }
 
-impl SbpLogger {
-    pub fn new_sbp<P: AsRef<Path>>(filepath: P) -> Result<SbpLogger> {
-        Ok(SbpLogger::Sbp(File::create(filepath)?))
+impl<W: Write> SbpLogger<W> {
+    pub fn new(logger: SbpLogging, out: W, path: Option<PathBuf>) -> Self {
+        Self { logger, out, path }
     }
```
pcrumley, line 62: I don't love this api design where the caller to the constructor has to path both writer W & Optional PathBuf, and it is up to the caller to know if `PathBuf` is some, we will require that file to exist whenever `serialize` is called.

what's to stop a person from calling something that looks like
```rust 
    let f = File::create("foo.txt")?;
    let mut f = BufWriter::new(f);
    let logger = SbpLogger::new(SbpLogging::SBP, f, None);
}
```
I don't really have a great suggestion here, but I think it may be better to design it as follows:

```rust
pub enum LoggingOutput<W: Write> {
    File { out: W, path: PathBuf } 
}

#[derive(Debug)]
pub struct SbpLogger<W: Write> {
    logger: SbpLogging,
    out: W,
    path: Option<PathBuf>,
}

impl<W: Write> SbpLogger<W> {
    pub fn new(logger: SbpLogging,  output: LoggingOutput) -> Self {
        match output {
            LoggingOutput:File {out, path} => Self { logger, out, path: Some(path} },
        }
    }
```

You might argue YAGNI, with the enum of one type. However, I think if you aren't going to need an enum here, I don't understand why you would need the new fn to take a generic writer & an optional pathbuf.

console_backend/src/output.rs
pcrumley: I think using something like [notify](https://docs.rs/notify/5.0.0/notify/) or [watchexec](https://docs.rs/watchexec/latest/watchexec/) is a good idea but may require a bigger refactor than this bug necessitates. At the same time I don't think this check actually prevents all the bugs you are trying to prevent here, and it may create some new ones. You are just checking if the path exists, which could happen if someone deleted the file and then recreated it, this check would pass, but the write_all path would drop.

Another issue is that if you rename the file with `mv` the code before this change will keep writing to the renamed file, whereas this change will cause an unnecessary error imo. 

Maybe these are tradeoffs that are ok to make, I don't have the context here and the jira tickets are minimal

console_backend/src/common_constants.rs
```diff
@@ -273,8 +273,8 @@ pub enum Keys {
     NEW_INS_CONFIRMATON,
     #[strum(serialize = "ANTENNA_STATUS")]
     ANTENNA_STATUS,
-    #[strum(serialize = "RECORDING_DURATION_SEC")]
-    RECORDING_DURATION_SEC,
+    #[strum(serialize = "RECORDING_START_TIME")]
+    RECORDING_START_TIME,
```
pcrumley, line 277: hard to know if this breaking change is safe.

pcrumley (commented):
Let my comments inline. FYI, the tests are failing comparing `Number(45)` to `Number(45.0)`

console_backend/src/output.rs
adrian-kong: yea, this would address file deletion but file moves and instantly recreated would be kinda edge case (can fix by just ensuring metadata / size of file path before writing as well - not sure how much performance that is tho) 🤔

console_backend/src/output.rs
adrian-kong: hmm yeah, i wasn't sure if abstraction is even needed here since idk if we would ever want to emit this data anywhere else - uploading ? or stdout or something? (which might also introduce some funky logic as to how we want to reflect # of bytes written or check the success of those) so kinda went in between

this just kinda explicitly state a link for the write to the pathbufs to address the UI not being updated (oOoOo spooky inodes 👻)

but if we do, maybe just have a callback for specific logger types would be best instead of a pathbuf to allow more flexibility 🤔

hmm maybe file lock would be nice or 
maybe best to just keep this simply as sbp file logger with pathbuf without doing funky `impl Write` stuff

console_backend/src/common_constants.rs
adrian-kong: oops yea good catch, (this one isn't related to this PR), but im pretty sure common_constants are autogenerated from python 

which is [here](https://github.com/swift-nav/swift-toolbox/blob/b9b8d3e43cd8c391be31a00296b311756377a799/swiftnav_console/constants.py#L151)

which should be OK, since we know explicitly where these capnproto packets are sent and handled

console_backend/src/output.rs
pcrumley: that's fine but then I would also remove the option on the `PathBuf`

adrian-kong:
will merge this back into #862 to finalize the ticket,

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3869626032)

PR #886, opened 2022-12-29
disconnecting bug [CPP-847]

adrian-kong:
atm just debugging what the bug actually is 🤔 

seems like RealTimeIter is sleeping for a long time which makes the rest of the messages not being processed, as well as unable to disconnect / play another file.

will need to look into this more.
but i would imagine the thread::sleep is actually overlapping and not being disconnected when requested.


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3797933845)

PR #888, opened 2023-01-03
pass and update current index [CPP-852]

adrian-kong:
this is only required to handle the initial case (can be modified by CLI).
currentIndex and checked is interactive and can only be changed via onActivated elsewhere hence we can just allow the interactive component to handle modifications of these variables.


swiftnav_console/logging_bar.py
```diff
@@ -97,6 +99,7 @@ def __init__(self):
 
     @classmethod
     def post_data_update(cls, update_data: Dict[str, Any]) -> None:
+        update_data[Keys.SBP_LOGGING_FORMAT_INDEX] = update_data[Keys.SBP_LOGGING_LABELS].index(update_data[Keys.SBP_LOGGING_FORMAT])
```
adrian-kong, line 102: it might be better to have this posted in capnproto but we construct the string model here so will just leave it here...

resources/LoggingBar.qml
```diff
@@ -287,11 +286,14 @@ Rectangle {
...
-                csvLoggingButton.checked = loggingBarData.csv_logging;
+                if (sbpLoggingFormat.currentIndex == -1) {
+                    sbpLoggingFormat.currentIndex = loggingBarData.sbp_logging_format_index;
+                    sbpLoggingButton.checked = loggingBarData.sbp_logging;
+                    csvLoggingButton.checked = loggingBarData.csv_logging;
+                }
                 if (loggingBarData.recording_filename)
                     recordingFilenameText.editText = loggingBarData.recording_filename;
-                let recording = sbpLoggingButton.checked || csvLoggingButton.checked;
+                let recording = loggingBarData.sbp_logging || loggingBarData.csv_logging;
```
john-michaelburke, line 296: I might be out of the loop on this but I'm not sure we want to mix these up. What happens if you record sbp and csv at the same time?

john-michaelburke (approved):

PR #889, opened 2023-01-04
sleep suspends beyond disconnect [CPP-858]

adrian-kong:
Bug:
- When file is corrupted / cut out with a huge time delay, toolbox goes into sleep and unable to connect to other files when attempting to disconnect.

Previously:

https://user-images.githubusercontent.com/35755741/210666918-bcea5737-030d-4d07-a97f-22d776b400c3.mov

Currently:

https://user-images.githubusercontent.com/35755741/210666981-10fab215-7147-4ea0-a445-6abb2bf577c1.mov

### DIFF
--- a/console_backend/src/connection.rs
+++ b/console_backend/src/connection.rs
@@ -209,7 +209,7 @@ fn conn_manager_thd(
                     refresh_loggingbar(&client_sender, &shared_state);
                     shared_state.set_connection(ConnectionState::Disconnected, &client_sender);
                     refresh_connection_frontend(&client_sender, &shared_state);
-                    drop(pm_thd.take());
+                    join(&mut pm_thd);
                     info!("Disconnected successfully.");
                 }
             };
@@ -220,7 +220,7 @@ fn conn_manager_thd(
             .expect("panicking sending stop message to status bar");
         join(&mut status_thd);
         join(&mut reconnect_thd);
-        drop(pm_thd.take());
+        join(&mut pm_thd);
     })
 }
 
--- a/console_backend/src/connection.rs
+++ b/console_backend/src/connection.rs
@@ -219,8 +219,8 @@ fn conn_manager_thd(
         tx.send(false)
             .expect("panicking sending stop message to status bar");
         join(&mut status_thd);
-        join(&mut reconnect_thd);
         join(&mut pm_thd);
+        join(&mut reconnect_thd);
     })
 }
 
--- a/console_backend/src/connection.rs
+++ b/console_backend/src/connection.rs
@@ -209,6 +209,7 @@ fn conn_manager_thd(
                     refresh_loggingbar(&client_sender, &shared_state);
                     shared_state.set_connection(ConnectionState::Disconnected, &client_sender);
                     refresh_connection_frontend(&client_sender, &shared_state);
+                    drop(pm_thd.take());
                     info!("Disconnected successfully.");
                 }
             };
@@ -218,8 +219,8 @@ fn conn_manager_thd(
         tx.send(false)
             .expect("panicking sending stop message to status bar");
         join(&mut status_thd);
-        join(&mut pm_thd);
         join(&mut reconnect_thd);
+        drop(pm_thd.take());
     })
 }
 
@@ -253,6 +254,7 @@ fn process_messages_thd(
             shared_state.clone(),
             client_sender,
         );
+        // when no more messages
         if conn.is_file() {
             manager_msg.send(ConnectionManagerMsg::Disconnect);
             if conn.close_when_done() {
--- a/console_backend/src/connection.rs
+++ b/console_backend/src/connection.rs
@@ -209,7 +209,6 @@ fn conn_manager_thd(
                     refresh_loggingbar(&client_sender, &shared_state);
                     shared_state.set_connection(ConnectionState::Disconnected, &client_sender);
                     refresh_connection_frontend(&client_sender, &shared_state);
-                    join(&mut pm_thd);
                     info!("Disconnected successfully.");
                 }
             };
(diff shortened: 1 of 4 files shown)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3842428304)

console_backend/src/connection.rs
```diff
@@ -209,7 +209,6 @@ fn conn_manager_thd(
                     refresh_loggingbar(&client_sender, &shared_state);
                     shared_state.set_connection(ConnectionState::Disconnected, &client_sender);
                     refresh_connection_frontend(&client_sender, &shared_state);
-                    join(&mut pm_thd);
```
pcrumley, line 212: it it safe to just not block on joining the thread here? I guess i get why this solves the problem, but could the thread end up being killed in such a way to leave it in a bad state? Like maybe the solution is to have a graceful shutdown signal we can send to pm_thd.

I'm just trying to understand why this join may have been here in the first place with very little to no understanding of the code.

Thx.

adrian-kong: > Like maybe the solution is to have a graceful shutdown signal we can send to pm_thd.

hmm yeah was thinking that too 🤔 but this seemed like a simpler solution (if there weren't any side effects im not sure about...) 

maybe an even better solution is to do spinning wait instead of these micro sleeps (which would save some resources since we can do stuff while it 'sleeping'?)

john-michaelburke: Most of the disconnect / reconnect logic was framed around tcp and serial connections. We had a large struggle on Windows with the serial package we use but has since been fixed. I would highly suggest doing reliability testing on serial + tcp when making changes like this. I do believe we were experiencing a lockup of some sort when there was no join here but that may not be an issue anymore.

You should look at what the errorkind is and if it is not handled in that match statement above, you could add a block and instruct it not to connect. There is another case in which it may already be handled and we are instructing it to reconnect, this will be tricky because we are mixing up reconnect logic for file and whatever other connection type. It could be the case that errorkind is associated with a failure we want to reconnect with for tcp/serial but not for file. That could be a case for errorkind + connection type specific reconnection logic.

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3869744582)

console_backend/src/connection.rs
adrian-kong: huh just a rust question - is joining necessary or just for good practice?

process_message thread basically processes the messages, if we disconnect - we should no longer have any messages to process (StopToken should force the iterator to end so thats all good, it would just be become an idle thread) so it SHOULD hypothetically be free to ignore or drop the thread whenever we disconnect

🤔 replaced this with `drop(...)` to drop this immediately but intuitively feels like it should just be an idle thread so theoretically dont actually need to join it...

pcrumley: joining a `JoinHandle` is necessary to make sure it completes it work & actually cancels the thread. for a standard os thread, dropping the thread just drops the handle and leaks the thread: https://doc.rust-lang.org/std/thread/struct.JoinHandle.html

You could run into scenarios where the thread continues to process items as a zombie thread, or it wouldn't gracefully shutdown if killed by a different process. TBH I think adding `drop` here is more confusing than not having it.

I can see how it seems to work fine in manual testing, because if you simply leak the thread and it doesn't do anything, you're not really causing any issues.

john-michaelburke: As mentioned before with the windows serial issues, what would usually get stuck if we did not join here was this subroutine for grabbing settings:
https://github.com/swift-nav/libsettings-rs/blob/0c80b8c4f7954890a30991047946c137ff023422/src/client.rs#L138

In some cases, we were seeing it run for more than 30sec. So if the user started a connection, closed it, then attempted to reconnect to the same device while the previous connection was still completing that subroutine we would hit a deadlock. Not sure what the deadlock was from though potentially a lock on the serial port that would block a new connection from connecting to it :shrug:

console_backend/src/tabs/settings_tab.rs
```diff
@@ -277,27 +277,14 @@ impl SettingsTab {
     }
 
     fn import_success(&self) {
-        let mut builder = Builder::new_default();
-        let msg = builder.init_root::<crate::console_backend_capnp::message::Builder>();
```
adrian-kong, line 281: unrelated changes - was playing around with threads in settings and refactored this out into utils

console_backend/src/connection.rs
adrian-kong: * will have to split this ticket out to investigate ^^,

ended up keeping the joining of the thread as well using park_timeout as @pcrumley suggested 👍

console_backend/src/process_messages.rs
```diff
@@ -127,9 +127,9 @@ pub fn process_messages(
         }
     })
     .expect(PROCESS_MESSAGES_FAILURE);
-
     let err = messages.take_err();
     let handle = messages.into_handle();
+    handle.thread().unpark();
```
pcrumley, line 132: I feel like this is kind of a weird place to put the unpark, to me it made more sense after the stopToken is set (e.g. see the example in the rust std: https://doc.rust-lang.org/std/thread/fn.park.html#examples

However, I think this should be fine as long as the parking timeout we are awaking here isn't used to rate limit the `process_messages` function.

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3880920461)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3888531589)

pcrumley (approved):
lgtm

PR #891, opened 2023-01-06
test with lint enabled to see warning in CI while I set up linting lo…

pcrumley:
WIP


pcrumley:
closing because this problem is leading me in circles.

PR #897, opened 2023-01-10
settings not disabled when group has changed [CPP-782]

adrian-kong:
TLDR: previously iterated over existing settings which did not handle case where we attempted to set configs that werent set causing it to not disable ntrip / ethernet before setting.

---

This was tested by running
```
cargo run --bin swift-settings --tcp ... --import config.ini
```
modifying ntrip in config.ini file,

previously would result in 
```
writing ntrip debug False
Error: setting is not modifiable
```
caused by iterating over the current settings 
(since original did not have debug, it would assume the group was not changed)

--- 

the new changes iterate over the modified keys (and checking over the changes instead)

the added logging statements show which settings were toggled and can see now ntrip gets disabled prior to settings.

- (NOTE* this was left unchanged) using `--write` in swift-settings command will need to manually specify this procedure


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3898751326)

console_backend/src/tabs/settings_tab.rs
```diff
@@ -599,18 +602,17 @@ impl SettingsTab {
...
+            .unwrap_or_default();
+        // if is in new config
         let in_config = conf
             .section(Some(group))
             .and_then(|s| s.get(name))
             .is_some();
         self.write_setting(group, name, value)?;
         if !in_config {
-            Ok(Some(original))
+            Ok(original) // revert to original if not in new config
```
silverjam, line 615: How does this handle the case where there is no original and it's not in the group, does it set the values to the default?

silverjam:
@adrian-kong Can you add a note to the PR about how this was tested?

console_backend/src/tabs/settings_tab.rs
adrian-kong: when theres no original, it wont revert back to original i.e. it gets set
since we can't fetch the field.

this was because ntrip.enable wasn't found - which would turn ntrip to false by default (im assuming if its not found, it was false?)

adrian-kong:
> @adrian-kong Can you add a note to the PR about how this was tested?

added 👍

console_backend/src/tabs/settings_tab.rs
```diff
@@ -620,19 +622,17 @@ impl SettingsTab {
...
-            let old_value = match value.as_ref().map(|v| v.to_string()) {
-                Some(v) => v,
-                None => return Ok(true),
-            };
-            if new_value != old_value {
+        let groups = match guard.get_group(group) {
+            Some(v) => v,
+            None => return Ok(true), // update since current values are empty
+        };
+        for (name, new_value) in new_group.iter() {
```
adrian-kong, line 632: this was basically the one of the main issue.

it did not check ntrip.debug as 'old values' did not have this field.
and hence when we tried to toggle ntrip.debug, we didn't actually disable ntrip which would make the first error

console_backend/src/tabs/settings_tab.rs
```diff
@@ -599,18 +602,17 @@ impl SettingsTab {
...
             .settings
             .lock()
-            .get(group, name)?
-            .value
-            .as_ref()
-            .ok_or_else(|| anyhow!("{group}.{name} was none"))?
-            .clone();
+            .get(group, name)
+            .map(|x| x.value.clone())
+            .unwrap_or_default();
```
adrian-kong, line 607: this was the second issue, `ntrip.enable` field was not found in old values and would terminate here.
this now assumes the groups are off by default so we won't unset it

silverjam (approved):

PR #903, opened 2023-01-16
purge qmlformat and qmlls from dist [CPP-857]

adrian-kong:
removes:
- qmlls (qml language server https://doc.qt.io/qtcreator/creator-language-servers.html)
- qmlformat

total: 12 mb
from dist (linux / mac)

this was the underlying cause to macOS signing issue as they are not codesigned (refer to https://swift-nav.atlassian.net/browse/CPP-857?focusedCommentId=70192)

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -715,7 +715,7 @@ if eq ${os} windows
 elseif eq ${os} mac
     rm -r ./share
     rm -r ./lib/python3.9/site-packages/PySide6/Qt/libexec
-    bins = array qmllint lupdate lrelease scripts
+    bins = array qmllint lupdate lrelease scripts qmlls qmlformat
     for bin in ${bins}
         files = glob_array ./lib/python3.9/site-packages/PySide6/${bin}
         for bin in ${files}
@@ -739,7 +739,7 @@ elseif eq ${os} mac
 else
     rm -r ./share
 
-    bins = array rcc uic designer pyside6-lupdate Designer.app
+    bins = array rcc uic designer pyside6-lupdate Designer.app qmllint qmlls qmlformat
     for bin in ${bins}
         files = glob_array ./**/PySide6/${bin}
         for bin in ${files}


adrian-kong:
think this should reoslve the mac signing issue so can re-release 4.1.7 alongside #862 changes (nice to get merged in as well)

* assuming we dont need to ship qmlls and qmlformat. alternatively we can codesign them as well.

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3926685373)

pcrumley (approved):
This makes a lot of sense and you probably could delete that qml stuff from the windows build as well.

Regarding the 're-release strategy' i think that if you want to do a re-release, you should apply this commit to the release branch and no other commits, IMO this is the most you could change a thing without have to update the semver string. Even if it was just this, i still think you should probably increase a minor version. However, if you wanted to add a new feature/bugfix, like you suggested i think it is better to release as `4.1.8` and not `4.1.7`.

That said, i'm open to following the way versioning is traditionally done on this project, just my 2 cents.

adrian-kong:
> delete that qml stuff from the windows build as well

think windows deletes all qtbins, (not exactly sure but split this out to https://swift-nav.atlassian.net/browse/CPP-862 to do some investigating 🔍) 

> Regarding the 're-release strategy'

yea but since 4.1.7 failed to release (mightve been updating to pyside 6.4) was gonna sneak in these bug fix changes

pcrumley:
I was missing this context that 4.1.7 was never released. Since we dont have RC's or anything, yeah feel free to sneak some extra goodies in, just follow regular QA processes for releases :)

adrian-kong:
*will merge this in for releasing.

if any issues with removing qmlls and qmlformat* lmk! 👍

silverjam:
👍 -- official releases will go through our support portal, so we can typically just release whenever appropriate from the GitHub project

PR #911, opened 2023-01-17
solutions axes equal aspect [CPP-738]

pcrumley:
This code is ready for review. It takes fixes the x and y range so that the pixel distances are the same in the coordinate unit in both x and y direction, that is, the grid lines should always appear as square on the screen and circular paths will appear circular and not ellipsoidal.

*update* -- i cleaned up the code so all of the code that handles the apsect ratio correction is in a function called `fixAspectRatio`. It's a little finicky because you have to make sure you call `fixAspectRatio` at the right points. A possible workaround would be to call fixAspectRatio whenever the axes min or max changes, but that would have some performance overhead (4x calls when changing all 4 axes values). Right now i think it is correct, but I don't think this solution is super robust to refactoring.

Screenshot 2023-01-18 at 9 21 27 AM


resources/SolutionTabComponents/SolutionPositionTab.qml
```diff
@@ -135,8 +135,8 @@ Item {
                 onClicked: {
                     if (checked) {
                         solutionZoomAllButton.checked = false;
-                        y_axis_half = Utils.spanBetweenValues(solutionPositionXAxis.max, solutionPositionXAxis.min) / 2;
-                        x_axis_half = Utils.spanBetweenValues(solutionPositionYAxis.max, solutionPositionYAxis.min) / 2;
+                        x_axis_half = Utils.spanBetweenValues(solutionPositionXAxis.max, solutionPositionXAxis.min) / 2;
+                        y_axis_half = Utils.spanBetweenValues(solutionPositionYAxis.max, solutionPositionYAxis.min) / 2;
```
pcrumley, line 139: This was an unrelated bug in the code

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3951690603)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3954344212)

adrian-kong (approved):
looks good to me! 👍

adrian-kong:
might want to change the title of PR before merging*

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3981923288)

pcrumley:
@adrian-kong do you have any suggestions for a new name? maybe:

`patrick/equal-xy-grid-spacing`?

adrian-kong:
> @adrian-kong do you have any suggestions for a new name? 

was thinking to just have `solution axes equal aspect` would be ok so it would be shorter title, (if there was a ticket as well `solutions axes equal aspect [CPP-...]`)

PR #912, opened 2023-01-17
fixes skyplot labels [CPP-831]

adrian-kong:
- reverts breaking changes from https://github.com/swift-nav/swift-toolbox/pull/802
- fixes ReferenceError for visibility field

### DIFF
--- a/resources/TrackingTabComponents/TrackingSkyPlotTab.qml
+++ b/resources/TrackingTabComponents/TrackingSkyPlotTab.qml
@@ -302,7 +302,7 @@ Item {
                         Label {
                             color: 'black'
                             text: '` + labels[idx][jdx] + `'
-                            visible: (!polarChartWidthChanging && labelsVisible && seriesEntry.visible)
+                            visible: !polarChartWidthChanging
                             width: 20
                             height: 20
                             x: ` + pose.x + `
--- a/swiftnav_console/main.py
+++ b/swiftnav_console/main.py
@@ -452,7 +452,9 @@ class BackendMessageReceiver(QObject):  # pylint: disable=too-many-instance-attr
             elif m.which == Message.Union.TrackingSkyPlotStatus:
                 data = tracking_sky_plot_update()
                 data[Keys.SATS][:] = m.trackingSkyPlotStatus.sats
-                data[Keys.LABELS][:] = m.trackingSkyPlotStatus.labels
+                data[Keys.LABELS][:] = [
+                    list(m.trackingSkyPlotStatus.labels[idx]) for idx in range(len(m.trackingSkyPlotStatus.labels))
+                ]
                 TrackingSkyPlotPoints.post_data_update(data)
             elif m.which == Message.Union.ObservationStatus:
                 data = observation_update()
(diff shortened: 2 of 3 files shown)


resources/TrackingTabComponents/TrackingSkyPlotTab.qml
```diff
@@ -302,7 +302,7 @@ Item {
                         Label {
                             color: 'black'
                             text: '` + labels[idx][jdx] + `'
-                            visible: (!polarChartWidthChanging && labelsVisible && seriesEntry.visible)
+                            visible: !polarChartWidthChanging
```
adrian-kong, line 305: since label is destroyed on timer, we dont need these visibility fields

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3944272962)

adrian-kong:
frontend bench failing on CI fixes - can ignore

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/3947594415)

resources/TrackingTabComponents/TrackingSkyPlotTab.qml
pcrumley: i agree this can be removed but I think your reasoning might be wrong (the logic earlier in the control flow ensures that at this point in the code `labelsVisible`  & `seriesEntry.visible` are always `true`.

pcrumley (commented):
To be honest, it's not clear what the regression was, i guess it was type errors of the `m.trackingSkyPlkotStatus.labels` & in the iterator of `Keys.SATS`? 

If the unit test infra for the python code is in place, I think it might be worth thinking about how to add a few tests there, just because it should be relatively straightforward, and in general is always good practice when you hit a regression bug.

resources/TrackingTabComponents/TrackingSkyPlotTab.qml
adrian-kong: yea the main issue was inserting object calling seriesEntry field which isn't declared globally causing the ReferenceError. im assuming it wanted to only use one label each (but coords not updated) but theres timer to clean it up whenever spawned so it didn't really make sense. 

i.e. if we dont have a timer to clean up labels then we would need visbility to refer to the series

adrian-kong:
> To be honest, it's not clear what the regression was, i guess it was type errors of the `m.trackingSkyPlkotStatus.labels` & in the iterator of `Keys.SATS`?

yee that was case

> might be worth thinking about how to add a few tests there, just because it should be relatively straightforward

yea unit tests would be nice... not sure how to go about it tho might be kinda funky tho. i.e. testing for UI changes / interactivity,

can use like https://doc.qt.io/qtforpython/overviews/qttestlib-tutorial1-example.html

maybe worth dropping a ticket in the backlog since i dont think we have any tests setup for python

pcrumley:
Ok cool, if the infra isn't there on the UI side, scratch the unit testing bit. I do think it might be worth thinking about ways to add better testing to the UI side, but that is a bigger issue than this one PR.

pcrumley (approved):

PR #917, opened 2023-01-27
updates dependencies, pylints and clippy

adrian-kong:
updates dependencies, pylints and new rust clippies.

PR #930, opened 2023-02-06
remove loglevel enum [CPP-884]

adrian-kong:
just a small hack related to LogLevel#error being generated to conflict `Self::Error`, 
since we hotfixed this in the #917 by using text instead of enums (think that takes more bytes), this enum is no longer needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,11 @@ Either connect to a device via the GUI or command line and include the flag:
 ```
 This will save a `.pickle` file in your current working directory.
 
-**NOTE:** This recording is not intended to be compatible with other console versions.
+**NOTE:** Use of this recording is intended only for debugging the swift-console and not for
+long term storage of streams. There is no guarantee the recording will be compatible with
+other console versions, as the messaging format between backend and frontend has no guarantee
+of backwards or forwards compatibility
+
 Debugging internal messaging should be version specific and recording it should be tied
 to its version.
 
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ Either connect to a device via the GUI or command line and include the flag:
 ```
 This will save a `.pickle` file in your current working directory.
 
+**NOTE:** This recording is not intended to be compatible with other console versions.
+Debugging internal messaging should be version specific and recording it should be tied
+to its version.
+
 ### (Debug) Replaying a stream from the backend in frontend (with or without Rust).
 If you have recorded a capnp recording pickle file as shown in the previous step, now you
 can replay this file. If you already have the standard development environment set up, you
@@ -420,6 +424,10 @@ just to cross language barriers). Other formats like protobufs require a parsing
 phase -- capnproto also has very ergonomic Python bindings. Other formats such as
 [flatbuffers][] achieve similar goals as *capnproto* but do not have good
 Python support.
+
+**We do not intend on allowing backward compatibility** in internal message passing via
+capnproto. There may be releases that modify internal message structures and should 
+not be relied on even though these internal packets are being exposed.
 
 [capnproto]: https://capnproto.org/
 [flatbuffers]: https://google.github.io/flatbuffers/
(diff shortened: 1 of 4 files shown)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4109142987)

pcrumley (commented):
Are these sorts of changes save to take? Is it true that this enum is in no other struct fields, and we can guarantee we will never see it (from a 3rd party or from a persisted capnp) If it was protobuf I believe the correct way to handle this is to make it as deprecated and not remove anything.

adrian-kong:
Hmmm I'm pretty sure they aren't being used anywhere as I swapped them to strings 🤔 it was one of the reasons that I wanted to get rid of this since it made it much more confusing as to what is being used or not / debugging

BUT yea, some of the capnp changes would break previous capnp recordings which is really bad - wonder how often old capnp recordings are used tho

pcrumley:
Given that, I would prefer someone else were to review this PR. Int the future based on what i have seen of capnp best practices, they recommend changing the name from `foo` to `obsoleteFoo` with a type of `AnyPointer` if the field is deprecated.

silverjam:
> Given that, I would prefer someone else were to review this PR. Int the future based on what i have seen of capnp best practices, they recommend changing the name from `foo` to `obsoleteFoo` with a type of `AnyPointer` if the field is deprecated.

I think we should probably start doing this, especially if we are going to persist capnp recordings like `console_backend/tests/data/console-capnp-20220419-033358.pickle` -- in the past we've treated capnp as ephemeral and in process only.

So I think options are:
- Start maintained backwards compat with capnp defs
- Delete `console_backend/tests/data/console-capnp-20220419-033358.pickle` and make it clear in the docs that capnp is not intended to work between different revisions the app.

silverjam:
> BUT yea, some of the capnp changes would break previous capnp recordings which is really bad - wonder how often old capnp recordings are used tho

We have one pickle file that's save in the repo, but it's only used here https://github.com/swift-nav/swift-toolbox/blob/8acb46a2acc55b4e5d790127f1a3951ba5bd64de/utils/loop_generate_core.sh#L25

This isn't used in CI, so it's probably save to delete `console_backend/tests/data/console-capnp-20220419-033358.pickle` and just change `utils/loop_generate_core.sh` to take a pickle file that's passed on the command line.

pcrumley:
I see, my bet is the breaking change related to loglevel that would have invalidated old capnp was the previous change that switched the log levels for strings.

I thought end users could also save recordings, and it isn't just important if it is in CI/CD?

adrian-kong:
I thought since it's exposing internal message passing - we shouldn't support backwards compatibility between versions since that would make it very hard to change the internal structures for only debugging purposes - though that would mean you can't debug which version broke (but now thinking it wouldn't be easy to debug on different versions anyways).

README.md
```diff
@@ -79,6 +79,10 @@ Either connect to a device via the GUI or command line and include the flag:
 ```
 This will save a `.pickle` file in your current working directory.
 
+**NOTE:** This recording is not intended to be compatible with other console versions.
+Debugging internal messaging should be version specific and recording it should be tied
+to its version.
```
pcrumley, line 84: i like it better to put the reason here too:

```
**NOTE:** Use of this recorded is intended only debugging the swift-console and not for long term
storage of streams. There is no guarantee the recording will be compatible with other console versions, as the
messaging format between backend and frontend has no guarantee of backwards or forwards compatibility

Debugging internal messaging should be version specific and recording it should be tied to its version.
```

pcrumley (approved):

pcrumley:
I had some suggestions for language clarity in the readme but other than that i think it is gtg

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4139860398)

PR #935, opened 2023-02-13
google maps [CPP-899]

adrian-kong:
using google maps instead of OSM

added also colour and focus button 

https://user-images.githubusercontent.com/35755741/218876333-81324e3e-a4de-4094-b8ec-279e6c8cc254.mov

there seem to be a bug in maplibre-gl where focusing the point while zooming lags as compared to google maps. (open issue)

### DIFF
--- a/resources/web/map/index.html
+++ b/resources/web/map/index.html
@@ -3,13 +3,18 @@
 <head>
     <meta charset="utf-8"/>
     <meta name="viewport" content="initial-scale=1,maximum-scale=1,user-scalable=no"/>
-    <script type="text/javascript" src="http://unpkg.com/maplibre-gl@2.4.0/dist/maplibre-gl.js"></script>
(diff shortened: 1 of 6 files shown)


resources/web/map/index.html
```diff
@@ -3,13 +3,18 @@
...
 </head>
 <body>
+<label id="focus">
+    <input type="checkbox" id="focus-btn">
+    <span>Focus</span>
+</label>
 <div id="map"></div>
 <script type="text/javascript" src="js/trajectory.js"></script>
+<!--api key here-->
+<script src="https://maps.googleapis.com/maps/api/js?key=*********&callback=initMap" async
```
adrian-kong, line 16: hmmmm any good idea to hide this key in html 🤔 was thinkin maybe generating this html file on compile or somethin

adrian-kong: using mapbox instead.

PR #938, opened 2023-02-15
using mapbox map tiles

adrian-kong:
image


resources/web/map/js/trajectory.js
```diff
@@ -1,41 +1,20 @@
...
-};
-
 const style = {
     "version": 8,
     "sources": {
         "osm": {
             "type": "raster",
-            "tiles": ["https://a.tile.openstreetmap.org/{z}/{x}/{y}.png"],
+            // "tiles": ["https://a.tile.openstreetmap.org/{z}/{x}/{y}.png"],
+            "tiles": ["https://api.mapbox.com/v4/mapbox.satellite/{z}/{x}/{y}.png?access_token=pk.eyJ1Ijoic3dpZnQtYWRyaWFuIiwiYSI6ImNsZTN1MW82bDA2OGgzdXFvOWFuZTJlMnEifQ.9nR8m0C-B_ISNR4r4cMExw"],
```
adrian-kong, line 7: this API key is free account (think it should be safe to leave in source(?) or maybe bake this into build but might be weird for setting up locally)
 
think satellite view looks pretty good, lots of cool tiles we can choose from with maplibre-gl + free BUT it wont be 'out of the box' with the two buttons like googlemaps, (would probably need to find a plugin or add buttons to switch between)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4180005214)

adrian-kong:
https://user-images.githubusercontent.com/35755741/218949985-80f90a60-4dfd-4c86-a20b-e1f9a489308a.mov

couple of bugs can be seen above^^ but style switcher seems very nice.
(note, the bugs are not specific to this implementation, maplibre-gl is port of mapbox) * also quite laggy while focused when zooming compared to google maps (this is an open ticket).

adrian-kong:
https://user-images.githubusercontent.com/35755741/219228498-6cf23335-624a-4c09-b162-b09ebac56aa0.mov

cool colours + bug fix where points get cut off since paused.

resources/web/map/js/trajectory.js
```diff
@@ -1,72 +1,91 @@
...
+                map.panTo(pos);
+            } else if (focus) map.panTo(pos);
+            data[id].features.push({
+                type: 'Feature',
+                geometry: {
+                    type: 'Point',
+                    coordinates: pos
+                }
+            });
+            map.getSource(`route${id}`).setData(data[id]);
```
adrian-kong, line 88: TODO: `setData` not found when switching styles. need to persist source layers?

PR #943, opened 2023-02-22
skip invalid elevation angle [CPP-895]

adrian-kong:
for the file attached in ticket gives elevation angle of -91 which causes
```
thread '<unnamed>' panicked at 'attempt to subtract with overflow', console_backend/src/tabs/tracking_tab/tracking_sky_plot_tab.rs:87:25
```

since sbp docs say elevation is supposed to be (-90, 90), will just skip on these invalid values.


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4239887095)

adrian-kong:
ignore failing CI atm - codesigning QtWebEngine for maps does not exist in non map

joelynch (approved):
Looks dodgy

PR #953, opened 2023-03-06
mapping accuracy ellipse [CPP-906]

adrian-kong:


resources/web/map/js/trajectory.js
```diff
@@ -102,22 +132,20 @@ new QWebChannel(qt.webChannelTransport, (channel) => {
...
-            type: 'Feature',
-            geometry: {
-                type: 'Point',
-                coordinates: pos
-            }
-        });
-        if (!map) return
+    chn.recvPos.connect((id, lat, lng, hAcc, vAcc) => {
+        const pos = [lat, lng],
+            RADIUS_MIN = 0.0002,
```
adrian-kong, line 137: hard cap minimum to be 20cm.
rendering too small is difficult for zooming out.

console_backend/src/tabs/solution_tab/solution_position_tab.rs
```diff
@@ -424,6 +430,8 @@ impl SolutionPositionTab {
     pub fn handle_pos_llh(&mut self, msg: PosLLH) {
         self.last_pos_mode = msg.mode();
         let pos_llh_fields = msg.fields();
+        self.v_acc = pos_llh_fields.v_accuracy;
```
adrian-kong, line 433: is there a specific orientation / angle for this ellipse?

silverjam: Suggest asking on [#technical-questions](https://snav.slack.com/archives/C02USAD03) -- @ljbade do you know?

resources/web/map/js/trajectory.js
silverjam: Not sure I understand, if we go to a certain zoom, then the accuracy circles should probably just be removed rather than forcing them to be a particular size.  Or, we can establish a metric (based on mapping meters to pixels) that would dictate whether it would be visible or not.

adrian-kong: yea... rendering even smaller is extremely difficult to see on the map.

i think drongo has it so the size gets bigger if accuracy decreases. 
** (accuracy circle is the same point, not an additional overlay that gets drawn around the location)

silverjam: @dgburr Any feedback here?

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4349507857)

resources/web/map/js/trajectory.js
ljbade: agree, shouldn't cap the min size, it is better to just not show it if its too small to render

console_backend/src/tabs/solution_tab/solution_position_tab.rs
ljbade: When there is only a horizontal accuracy then the ellipse should be round. Worth noting horizontal means parallel to the ground plane, vertical means perpendicular to the ground plane.

Thus only horizontal accuracy would be visible in a top down map view.

resources/web/map/js/trajectory.js
```diff
@@ -102,22 +132,20 @@ new QWebChannel(qt.webChannelTransport, (channel) => {
...
+            // cap radius to be 20cm minimum, for visibility on map
+            rX = Math.max(RADIUS_MIN, hAcc / 1000),
```
ljbade, line 139: the map plot should be a circle with radius hAcc, the vAcc is not visible on a top down map view

ljbade (changes requested):
Only horizontal accuracy should be plotted on a top down map view

resources/web/map/js/trajectory.js
adrian-kong: ahhh so it should just be a circle

resources/web/map/js/trajectory.js
adrian-kong: adjusted

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4351074626)

resources/web/map/js/trajectory.js
dgburr: I can confirm that Drongo does not enforce a minimum cap either: https://github.com/swift-nav/ats/blob/master/drive-exe/swift/map/scene/render.py#L36

ljbade (approved):

resources/web/map/js/trajectory.js
```diff
@@ -81,35 +77,68 @@ function syncLayers() {
...
+
+/**
+ * Helper method to create elliptical geojson data
+ * @param center {[lng: number, lat: number]}
+ * @param rX horizontal radius in kilometers of ellipse
+ * @param rY vertical radius in kilometers of ellipse
+ * @param points optional number of points to render ellipse, higher for smoothness
+ * @return {{geometry: {coordinates: [][], type: string}, type: string}}
+ */
+function createGeoJsonEllipse(center, rX, rY, points = 32) {
```
ljbade, line 96: the ellipse code will be useful if in future we want to plot the `MSG_POS_LLH_ACC:` or `MSG_POS_LLH_COV:` messages which allow break down of the horizontal accuracy into North/South and East/West components

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4360053119)

PR #958, opened 2023-03-08
obfuscate mapbox token

silverjam:


adrian-kong (approved):
should we also add an additional env variable for directly supplying the token without decoding (easier for local environment debugging)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4360780682)

silverjam:
> should we also add an additional env variable for directly supplying the token without decoding (easier for local environment debugging)

Yeah, that would be great

PR #960, opened 2023-03-09
render protection level circles [CPP-907]

adrian-kong:
adds protection level cricles. breadcrumb trail

https://user-images.githubusercontent.com/35755741/224184059-63ade663-4b9e-49f8-9b50-77cfb2df53de.mov

### DIFF
--- a/resources/web/map/js/trajectory_raw.js
+++ b/resources/web/map/js/trajectory_raw.js
@@ -5,7 +5,7 @@ const LNG_KM = 111.320, LAT_KM = 110.574;
 
 function decode(r){var n=r,t=[0,10,13,34,38,92],e=new Uint8Array(1.75*n.length|0),f=0,o=0,a=0;function i(r){o|=(r<<=1)>>>a,8<=(a+=7)&&(e[f++]=o,o=r<<7-(a-=8)&255)}for(var u=0;u<n.length;u++){var c,d=n.charCodeAt(u);127<d?(7!=(c=d>>>8&7)&&i(t[c]),i(127&d)):i(d)}r=new Uint8Array(e,0,f);var s=new TextDecoder().decode(r);while (s.slice(-1)=="\x00") s=s.slice(0,-1); return s;}
 
-mapboxgl.accessToken = decode("@ACCESS_TOKEN@");
+mapboxgl.accessToken = "pk.eyJ1Ijoic3dpZnQtYWRyaWFuIiwiYSI6ImNsZTN1MW82bDA2OGgzdXFvOWFuZTJlMnEifQ.9nR8m0C-B_ISNR4r4cMExw";
 var map = new mapboxgl.Map({
     container: 'map',
     style: "mapbox://styles/mapbox/light-v11",
@@ -77,7 +77,8 @@ function setupLayers() {
             source: `route${i}`,
             paint: {
                 'fill-color': lines[i],
-                'fill-opacity': 0.3
+                'fill-opacity': 0.3,
+                'fill-outline-color': '#000000'
             }
         });
     }
--- a/resources/web/map/js/trajectory_raw.js
+++ b/resources/web/map/js/trajectory_raw.js
@@ -1,6 +1,7 @@
 import mapboxGlStyleSwitcher from 'https://cdn.skypack.dev/mapbox-gl-style-switcher';
 
 const lines = ["#FF0000", "#FF00FF", "#00FFFF", "#0000FF", "#00FF00", "#000000"];
+const LNG_KM = 111.320, LAT_KM = 110.574;
 
 function decode(r){var n=r,t=[0,10,13,34,38,92],e=new Uint8Array(1.75*n.length|0),f=0,o=0,a=0;function i(r){o|=(r<<=1)>>>a,8<=(a+=7)&&(e[f++]=o,o=r<<7-(a-=8)&255)}for(var u=0;u<n.length;u++){var c,d=n.charCodeAt(u);127<d?(7!=(c=d>>>8&7)&&i(t[c]),i(127&d)):i(d)}r=new Uint8Array(e,0,f);var s=new TextDecoder().decode(r);while (s.slice(-1)=="\x00") s=s.slice(0,-1); return s;}
 
@@ -14,6 +15,7 @@ var map = new mapboxgl.Map({
 
 var focusCurrent = false;
 var startMarker = null;
+var currentMarker = null;
 
 class FocusToggle {
     onAdd(map) {
@@ -77,15 +79,37 @@ function setupLayers() {
             }
         });
     }
+    if (map.getSource(`prot`) != null) return;
+    map.addSource(`prot`, {
+        type: 'geojson',
+        cluster: false,
+        data: {
+            type: 'FeatureCollection',
+            features: []
+        }
+    })
+    map.addLayer({
+        id: `prot`,
+        type: 'fill',
+        source: `prot`,
+        paint: {
+            'fill-color': "#00FF00",
+            'fill-opacity': 0.5
+        }
+    });
 }
 
 function syncLayers() {
+    // sync route datas with stored points
     for (let i = 0; i < lines.length; i++) {
         map.getSource(`route${i}`).setData(data[i]);
     }
+    // clear protection, since its only one point and temporary
+    map.getSource('prot').setData({
+        type: 'FeatureCollection',
+        features: []
+    });
 }
-
-const LNG_KM = 111.320, LAT_KM = 110.574;
 
 /**
  * Helper method to create elliptical geojson data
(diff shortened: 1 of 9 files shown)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4379513527)

resources/web/map/js/trajectory_raw.js
```diff
@@ -77,16 +79,38 @@ function setupLayers() {
             }
         });
     }
+    if (map.getSource(`prot`) != null) return;
+    map.addSource(`prot`, {
+        type: 'geojson',
+        cluster: false,
+        data: {
+            type: 'FeatureCollection',
```
adrian-kong, line 87: we probably dont need a feature collection here>
since its only one point (current) on drongo

resources/web/map/js/trajectory_raw.js
```diff
@@ -77,16 +79,38 @@ function setupLayers() {
...
+            features: []
+        }
+    })
+    map.addLayer({
+        id: `prot`,
+        type: 'fill',
+        source: `prot`,
+        paint: {
+            'fill-color': "#00FF00",
```
adrian-kong, line 96: 00ff00 is also same colour as id=5 mode for line points. (but opacity is a bit higher)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4379778925)

resources/web/map/js/trajectory_raw.js
adrian-kong: @dgburr any thoughts on the colouring of this 🤔 (ok to have green for both?)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4410839068)

resources/web/map/js/trajectory_raw.js
```diff
@@ -131,19 +194,37 @@ new QWebChannel(qt.webChannelTransport, (channel) => {
...
+        if (map.getSource('breadcrumb')) syncCrumbCoords();
+    })
+
+    chn.protPos.connect((lat, lng, hpl) => {
+        const pos = [lng, lat], rX = hpl / 100_000; // hpl in cm, convert to km
+        if (!map) return;
+        let src = map.getSource(`prot`);
+        if (src) src.setData({
+            type: 'FeatureCollection',
+            features: [createGeoJsonEllipse(pos, rX, rX)]
```
john-michaelburke, line 226: I'm not familiar enough with this message type in the sbp protocol but is there ever a case when these values should be reset to 0 or a maximum duration before these need to get reset?

adrian-kong: hmmmmm oh yeah not quite sure when it should get reset
with the protocol also not sure whether its always sent along with the LLH msg

(but it will just persist if theres no new protection lvl msg, itll be like last seen position with protection lvl)
^^ will have to do more digging into drongo

resources/web/map/js/trajectory_raw.js
adrian-kong: ahh, hmmm then we'll have to figure some different ways to distinguish accuracy points and prot lvl 🤔 - 
toolbox currently uses colours to encode different points
basically took the colour scheme 

image

and encoded into the accuracy points which `RTK Fixed` also shares green colour as well

resources/web/map/js/trajectory_raw.js
john-michaelburke: Maybe you could kick off a timer event of some sort and reset to zero after 1sec has elapsed. I imagine we would be doing something similar in production. What are your thoughts @dgburr?

adrian-kong: ok yea i think it seems to have the same functionality compared to drongo

sbp file in: https://swift-nav.atlassian.net/browse/CPP-915

john-michaelburke: Fair enough, yeah I guess if drongo is reusing llh then it is probably acceptable :shrug:

resources/web/map/js/trajectory_raw.js
```diff
@@ -1,10 +1,11 @@
 import mapboxGlStyleSwitcher from 'https://cdn.skypack.dev/mapbox-gl-style-switcher';
 
 const lines = ["#FF0000", "#FF00FF", "#00FFFF", "#0000FF", "#00FF00", "#000000"];
+const LNG_KM = 111.320, LAT_KM = 110.574;
 
 function decode(r){var n=r,t=[0,10,13,34,38,92],e=new Uint8Array(1.75*n.length|0),f=0,o=0,a=0;function i(r){o|=(r<<=1)>>>a,8<=(a+=7)&&(e[f++]=o,o=r<<7-(a-=8)&255)}for(var u=0;u<n.length;u++){var c,d=n.charCodeAt(u);127<d?(7!=(c=d>>>8&7)&&i(t[c]),i(127&d)):i(d)}r=new Uint8Array(e,0,f);var s=new TextDecoder().decode(r);while (s.slice(-1)=="\x00") s=s.slice(0,-1); return s;}
 
-mapboxgl.accessToken = decode("@ACCESS_TOKEN@");
+mapboxgl.accessToken = "pk.eyJ1Ijoic3dpZnQtYWRyaWFuIiwiYSI6ImNsZTN1MW82bDA2OGgzdXFvOWFuZTJlMnEifQ.9nR8m0C-B_ISNR4r4cMExw";
```
adrian-kong, line 8: mapboxgl.accessToken = decode("@ACCESS_TOKEN@");

adrian-kong, line 8:
```suggestion
mapboxgl.accessToken = decode("@ACCESS_TOKEN@");	
```

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4431897633)

silverjam:
Seeing an error when attempting to run this branch: `qml: [MAP LOG] @45: Uncaught TypeError: map.addControl is not a function`

adrian-kong:
> Seeing an error when attempting to run this branch: `qml: [MAP LOG] @45: Uncaught TypeError: map.addControl is not a function`

hmm i cant get this error 🤔 , could you send the `trajectory_raw.js` file

resources/web/map/js/trajectory_raw.js
```diff
@@ -1,10 +1,11 @@
 import mapboxGlStyleSwitcher from 'https://cdn.skypack.dev/mapbox-gl-style-switcher';
 
 const lines = ["#FF0000", "#FF00FF", "#00FFFF", "#0000FF", "#00FF00", "#000000"];
+const LNG_KM = 111.320, LAT_KM = 110.574;
 
 function decode(r){var n=r,t=[0,10,13,34,38,92],e=new Uint8Array(1.75*n.length|0),f=0,o=0,a=0;function i(r){o|=(r<<=1)>>>a,8<=(a+=7)&&(e[f++]=o,o=r<<7-(a-=8)&255)}for(var u=0;u<n.length;u++){var c,d=n.charCodeAt(u);127<d?(7!=(c=d>>>8&7)&&i(t[c]),i(127&d)):i(d)}r=new Uint8Array(e,0,f);var s=new TextDecoder().decode(r);while (s.slice(-1)=="\x00") s=s.slice(0,-1); return s;}
 
-mapboxgl.accessToken = decode("@ACCESS_TOKEN@");
+mapboxgl.accessToken = decode("@ACCESS_TOKEN@");
```
adrian-kong, line 8:
```suggestion
mapboxgl.accessToken = decode("@ACCESS_TOKEN@");
```

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4443150226)

resources/web/map/js/trajectory_raw.js
silverjam: > @adrian-kong: are you asking whether PL circles and EHPE circles can have the same colour, or are you asking whether PL circles and their connecting lines can have the same colour?
> 
> IMHO it should work the same way as the Drongo defaults:
> 
> 1. Estimated accuracy ellipses are orange
> 2. Black lines connect estimated accuracy circles
> 3. PL ellipses are green

I think we need to decide on a different color scheme, part of the new console work was to change the color scheme that the console uses to line up with other tools: https://swift-nav.atlassian.net/browse/CPP-449

resources/web/map/js/trajectory_raw.js
dgburr: If I understand correctly, there are two questions here:

Q. Should the centre of the EHPE ellipses be the same as the centre of the Protection Level ellipses?  
A: Yes.

Q. When should the Protection Level ellipses disappear?
A: The way the SBP protocol works is that it is necessary to combine data from multiple SBP messages with the same timestamp to determine the complete state at any given moment.  In the source code for Drongo and Woodpecker, I use the term "epoch" to refer to this concept. One epoch can contain 0 or 1 HPL values (depending upon Starling was configured was configured to output MSG_PROTECTION_LEVEL messages).  If a HPL value is present in an epoch, then the PL circle should remain on the screen for as long as the epoch is being displayed.  Typically this would mean that it stays on the screen until all data for the next epoch is received, after which the old ellipse is removed and a new ellipse is drawn.

resources/web/map/js/trajectory_raw.js
adrian-kong: image

Here's the new colour scheme just to avoid drastic changes,
I've added outlines to accuracy circle and different shade of green.

PR #968, opened 2023-03-14
set validator locale to C [CPP-886]

adrian-kong:
https://doc.qt.io/qt-6/qml-qtquick-doublevalidator.html#locale-prop
https://doc.qt.io/qt-6/qml-qtqml-qt.html#locale-method

enforce C locale to prevent typing '1,234' vs '1.234' on different keyboard layouts

### DIFF
--- a/resources/SettingsTabComponents/SettingsPane.qml
+++ b/resources/SettingsTabComponents/SettingsPane.qml
@@ -405,11 +405,9 @@ Rectangle {
     }
 
     floatValidator: DoubleValidator {
-        locale: "C"
     }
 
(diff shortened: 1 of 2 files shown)


adrian-kong:
https://swift-nav.atlassian.net/browse/CPP-886?focusedCommentId=72900

not sure if this is an issue required to fix but this should standardize locale as '1.234'

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4414235201)

silverjam (approved):

PR #971, opened 2023-03-15
add --disable-map cli option [CPP-911]

adrian-kong:
apparently tab loads are lazy.
so we can basically just remove the tab from layout.
tested with javascript logging 'loaded' msg that only pops up whenever you click into the map.

also means all the concerns about performance previously can be disregarded. is actually pretty good not as bad as imagined.


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4422564419)

ljbade (approved):

PR #972, opened 2023-03-17
replace timers with functions [CPP-872]

adrian-kong:
replaces timers to update whenever received new data

1) means no more batched updates based on timer (may affect performance without strictly enforcing refresh rate)
2) means more realtime update


resources/SettingsTabComponents/SettingsTable.qml
```diff
@@ -121,6 +121,46 @@ Rectangle {
...
+                    } else {
+                        if (!showExpert && !groupHasNonExpertSetting(entries, idx)) {
+                            offset++;
+                            return;
+                        }
+                        new_row = headerRow(entry);
+                    }
+                    rowOffsets[idx - offset] = idx;
+                    tableView.model.setRow(idx - offset, new_row);
+            });
```
adrian-kong, line 161:
```suggestion
                });
```

resources/AdvancedTabComponents/AdvancedImuTab.qml
```diff
@@ -152,44 +181,6 @@ Item {
...
             }
-
-            Timer {
-                id: advancedImuTimer
-
-                interval: Utils.hzToMilliseconds(Globals.currentRefreshRate)
-                running: true
-                repeat: true
-                onTriggered: {
-                    if (!advancedImuTab.visible)
```
adrian-kong, line 163: can remove `.visible` checks since capnproto is paused in backend

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4464883163)

pcrumley (changes requested):
In general, I think this is a great idea and we should take this update.

However I have two main concerns: 

1) There is no explicit rate-limiting of calling the update function. I guess it probably is fine because the rust backend is rate limited at some level, but I'm not sure if we want to explicitly do something here at the front end, otherwise there could be perf issues.

2) I haven't carefully looked at all of the code you refactored, but the timer would be called regardless if the data are updated. It's possible that people could have snuck code in there that may have handled things like window/zoom resizing or something that only had to do with frontend changes. Did you carefully audit the timer codes to confirm this wasn't the case? or if there were such things, they should be attached to the slot where the signal for that change would be emitted.

adrian-kong:
> 1. There is no explicit rate-limiting of calling the update function.

not too sure on this but im guessing they have their own FPS cap for repainting 

> It's possible that people could have snuck code in there that may have handled things like window/zoom resizing or something that only had to do with frontend changes. 

oh yea good point. will look into this

resources/TrackingTabComponents/TrackingSkyPlotTab.qml
```diff
@@ -33,11 +33,61 @@ Item {
 
     property alias all_series: trackingSkyPlotPoints.all_series
     property var series: []
+    property var labelObjs: []
     property bool labelsVisible: labelsVisibleCheckBox.checked
     property bool polarChartWidthChanging: false
 
+    function clearLabels() {
```
adrian-kong, line 40: @pcrumley yeah it actually introduced a bug (no longer getting as frequent repainting of labels) so the timer sleep was making the labels flash.

this should be fixed now, also less refreshing now

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4485051930)

silverjam (changes requested):
How will this affect things like the refresh rate switch?  How well does this run on the surface tablet that you have for testing?

https://github.com/swift-nav/swift-toolbox/blob/6acfd17629806a5ce1ca3a10c65d66b99886aa87/swiftnav_console/main.py#L721

silverjam:
Showing a pretty big jump in CPU usage on my laptop.

Old
Screenshot 2023-03-21 192827

New
Screenshot 2023-03-21 193536

adrian-kong:
> Showing a pretty big jump in CPU usage on my laptop.
> 
> Old Screenshot 2023-03-21 192827
> 
> New Screenshot 2023-03-21 193536

are you in the Maps tab? seems like QtWeebEngineProcess is taking up most the CPU

silverjam:
> > How will this affect things like the refresh rate switch?
> 
> played around refresh rate switch doesnt do much after these changes but this does enforce repainting on data tho which is pretty nice, would think theres a way to cap FPS on pyside instead maybe?

Yes, because the refresh rate switch is made useless by this PR -- we should remove it or rethink how it works.

The way this worked in the previous console was to just record the last update time and skip an update if it was "too soon" -- so, effectively let the data stream drive updates, but painting would be skipped if it was exceeding the configured refresh rate.

Example: https://github.com/swift-nav/piksi_tools/blob/5b723eb0bf9af28f26b6190c9e3dfa5551da7123/piksi_tools/console/solution_view.py#L439

> > How well does this run on the surface tablet that you have for testing?
> 
> haven't tested on windows but for mac, the settings tab is much better where previously i was unable to scroll on

Maybe we should think about making a more targeted update(s) first then, like just removing the timer from the settings view?

silverjam:
More informal benchmarks --

Latest main branch: + have not clicked into the map tab + tracking view full ==>

image

This branch: + have not clicked into the map tab + tracking view full ==>

image

---

After running for a while the main branch does start to look more like this branch, so not sure there's a clear perf story here, but definitely seems like the web process is doing a lot more than it should be (especially if it's not loaded):

image

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4486189295)

adrian-kong:
removes refresh rate, since this makes the option redundant.
also see ticket
https://swift-nav.atlassian.net/browse/CPP-806

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4526982875)

adrian-kong:
@silverjam this seems to have roughly identical ~30% cpu (+- 1%) performance for cpu utilization after https://github.com/swift-nav/swift-toolbox/pull/989
compared to main branch on mac

silverjam (approved):

silverjam:
@pcrumley Please approved if your concerns have been addressed.

pcrumley (approved):

resources/web/map/js/trajectory_raw.js
```diff
@@ -5,7 +5,7 @@ const LNG_KM = 111.320, LAT_KM = 110.574;
 
 function decode(r){var n=r,t=[0,10,13,34,38,92],e=new Uint8Array(1.75*n.length|0),f=0,o=0,a=0;function i(r){o|=(r<<=1)>>>a,8<=(a+=7)&&(e[f++]=o,o=r<<7-(a-=8)&255)}for(var u=0;u<n.length;u++){var c,d=n.charCodeAt(u);127<d?(7!=(c=d>>>8&7)&&i(t[c]),i(127&d)):i(d)}r=new Uint8Array(e,0,f);var s=new TextDecoder().decode(r);while (s.slice(-1)=="\x00") s=s.slice(0,-1); return s;}
 
-mapboxgl.accessToken = decode("@ACCESS_TOKEN@");
+mapboxgl.accessToken = "pk.eyJ1Ijoic3dpZnQtYWRyaWFuIiwiYSI6ImNsZTN1MW82bDA2OGgzdXFvOWFuZTJlMnEifQ.9nR8m0C-B_ISNR4r4cMExw";
```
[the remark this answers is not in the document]
adrian-kong: oops... yea i agree (really tedious to remember to make sure its not committed)
https://swift-nav.atlassian.net/browse/CPP-919
just dropped a ticket here:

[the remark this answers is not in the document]
pcrumley: @adrian-kong I'd really recommend a git hook here. It's not a long term solution because each team member would have to set up the hooks (they aren't automatically tracked as part of the git repo).

but if you create a file called commit-msg in your `path/to/swift-toolbox/.git/hooks/` folder that looks like this:
```
#!/bin/sh

# This catches matboxgl accessTokens
if git grep -E "mapboxgl.accessToken[ ]*=[ ]*\"[a-zA-Z0-9_.\-]{96}\""; then
    echo "Detected mapbox access token was comitted"
    exit 2
fi
```
this will not happen again.

PR #974, opened 2023-03-20
improve map performance [CPP-916]

adrian-kong:
https://docs.mapbox.com/help/troubleshooting/mapbox-gl-js-performance/

- disables sending metrics to mapbox
- enables optimize base styles
- enables attribute use desktop opengl
- half required points to draw ellipse (might be better to scale this w.r.t distance on map since small circles looks ok rendering ~ 8 points)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4464899614)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4465864245)

resources/web/map/js/trajectory_raw.js
```diff
@@ -159,12 +160,13 @@ function syncLayers() {
  * @param points optional number of points to render ellipse, higher for smoothness
```
pcrumley, line 160: you need to move this doc string to be an inline comment about the `let points = 16; line.` OR undo the refactor of the optional param to a function.

pcrumley (changes requested):
This a pretty difficult thing to review other than just a rubber stamp, but it looks fine? 

marking as request changes because you need to update the documentation or undo a refactor

pcrumley (approved):
I kind of disagree with decision to remove the comments, i think putting it as an inline comment makes more sense but i don't think it should block merge.

adrian-kong:
> I kind of disagree with decision to remove the comments, i think putting it as an inline comment makes more sense but i don't think it should block merge.

think it might be changed in later PRs if the optimal # of points for zoom levels is figured out
but for now might just leave default 16 pts => 20 degrees

PR #986, opened 2023-03-23
regex validator to allow both dot and comma [CPP-886]

adrian-kong:
it seems both , and . are wanted. (not quite sure what the issue is but i'm assuming sometimes one doesn't work? - unable to reproduce the bug but heres a fix to allow for both to use as decimals)

### DIFF
--- a/resources/SettingsTabComponents/SettingsPane.qml
+++ b/resources/SettingsTabComponents/SettingsPane.qml
@@ -365,7 +365,10 @@ Rectangle {
             anchors.centerIn: parent
             anchors.verticalCenterOffset: 5
             onEditingFinished: {
+                if (settingType === "float" || settingType === "double")
+                    text = text.replace(",", ".");
                 backend_request_broker.settings_write_request(settingGroup, settingName, text);
+                console.log(text);
             }
             validator: {
                 if (settingType === "integer")
@@ -404,7 +407,8 @@ Rectangle {
         }
     }
 
-    floatValidator: DoubleValidator {
+    floatValidator: RegularExpressionValidator {
+        regularExpression: /[-+]?[0-9]*[.,]?[0-9]+/
     }
 
     intValidator: IntValidator {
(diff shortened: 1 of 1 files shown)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4497202953)

resources/SettingsTabComponents/SettingsPane.qml
```diff
@@ -365,6 +365,8 @@ Rectangle {
             anchors.centerIn: parent
             anchors.verticalCenterOffset: 5
             onEditingFinished: {
+                if (settingType === "float" || settingType === "double")
+                    text = text.replace(",", ".");
```
pcrumley, line 369: is this to support European style number delimiters? Is that what was the source of the original bug? I only skimmed the conversation in Jira.

adrian-kong: i'm not certain the bug (unable to change values in settings table) was due to the validators but this is just to support both `.` and `,`, how double can be inputted / ensure that its sending the correct setting writes to eliminate possibility for validator error(?) (this removes scientific notation as well, not sure how much that is used)

silverjam:
Are we sure that we want to support this?  If we're worried that `DoubleValidator` changes it's behavior based on the locale of the Windows system then perhaps we should just convert "," or "." and avoid changing the validator?

silverjam:
> > Are we sure that we want to support this? If we're worried that `DoubleValidator` changes it's behavior based on the locale of the Windows system then perhaps we should just convert "," or "." and avoid changing the validator?
> 
> ya this was a thing i wasnt sure how to go around how to go around because it would be weird not using doublevalidator to validate doubles, https://stackoverflow.com/questions/62148177/doublevalidator-allows-to-type-point-and-comma-symbols-in-textinput here's the issue i was reading and a couple more where they just used regexvalidator instead of doublevalidator its pretty old not sure if its replicable
> 
> this does remove scientific notation tho
> 
> BUT it does restrict the inputs tho which is nice since doublevalidator allows both , . as input but sometimes doesnt process the , (see in https://swift-nav.atlassian.net/browse/CPP-886?focusedCommentId=73858)

Thanks, I would take some time in the future to provide more context in pull request.  Perhaps we should setup a pull request template for this repo.

PR #987, opened 2023-03-24
add loader [CPP-920]

adrian-kong:
https://user-images.githubusercontent.com/35755741/229478660-fe1a4d20-4299-48a3-b7c1-bcbe848a35d2.mp4

sped up by x4.
loader shows when clicking refresh button

ideally might change this to not use qml loaders, wanting to make this a single bar and do some lerping so it will look much cleaner (?)

trying to figure out how to best design this but sucks cant hotswap changes directly on console 🤮


adrian-kong:
https://user-images.githubusercontent.com/35755741/229647105-40ccaf08-4ff1-4fa1-8e15-99ad30dc559e.mov

think this looks way better. (will have to address the case when tcp connected but no settings etc... maybe just make this bar only on refresh action + write action?)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4602533678)

resources/SettingsTabComponents/SettingsTable.qml
```diff
@@ -241,10 +254,7 @@ Rectangle {
             model: TableModel {
                 id: tableModel
 
-                rows: [{
-                        "Name": "",
-                        "Value": ""
-                    }]
+                rows: []
```
silverjam, line 257: Why do we need to remove this?

adrian-kong: this creates an empty row here:
image

adrian-kong: removing it makes it like this:
image

silverjam: 👍

silverjam (approved):

PR #989, opened 2023-03-26
disabled on default, swap to --enable-map

adrian-kong:
as per memory / cpu usage request, disables map by default

### DIFF
--- a/swiftnav_console/main.py
+++ b/swiftnav_console/main.py
@@ -258,7 +258,7 @@ TAB_LAYOUT = {
 
 capnp.remove_import_hook()  # pylint: disable=no-member
 
-MAP_ENABLED = False
+MAP_ENABLED = [False]
 
 
 class BackendMessageReceiver(QObject):  # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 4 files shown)


swiftnav_console/main.py
```diff
@@ -367,9 +369,11 @@ def _process_message_buffer(self, buffer):
                 data[Keys.LON_MIN] = m.solutionPositionStatus.lonMin
                 data[Keys.AVAILABLE_UNITS][:] = m.solutionPositionStatus.availableUnits
                 data[Keys.SOLUTION_LINE] = m.solutionPositionStatus.lineData
-                SolutionMap.send_pos(m.solutionPositionStatus)
+
+                if MAP_ENABLED:
+                    SolutionMap.send_pos(m.solutionPositionStatus)
```
adrian-kong, line 374: i dont really like this hacky business but:

the cpu / mem issue lies in sending position in the background
- this was done because sending an array of points via signals was more complex (i.e. sending array of pts then sending one by one updates, caching a bunch of pts etc...)
- but an issue is webengineview linked to the webchannel, (if theres a way to somehow decouple these classes? - sending points in the background should be fine as long as its not sending to webchannel)

adrian-kong:
these changes seems to put qtwebengineprocess to 0%

silverjam (approved):

PR #1002, opened 2023-03-31
python: try bumping to Python 3.11 again [CPP-882]

silverjam:
Try bumping to Python 3.11 one more time.

Related PR: https://github.com/swift-nav/swift-toolbox-ci/pull/9

Test build of installers here: https://github.com/swift-nav/swift-toolbox/releases/tag/v4.0.9-py311

Jira: https://swift-nav.atlassian.net/browse/CPP-882

PR #1004, opened 2023-04-02
use switch cases

adrian-kong:


adrian-kong:
using this branch on solution pos
Screenshot 2023-04-03 at 9 39 18 am

using silverjam/py311
Screenshot 2023-04-03 at 9 44 13 am

seems like switch is slower (?)
might be better readability but not really i guess. can disregard this change

adrian-kong:
#1002

silverjam:
Maybe a diction lookup to some functions would be faster?

adrian-kong:
hmm yea maybe - was hoping the python bump can use some cool match statements but nvm.
might be too tiny to measure performance changes on if elses to using map lookups, will just close this PR :(

PR #1007, opened 2023-04-03
fix sticky settings value bug [CPP-923]

adrian-kong:
some how `text` gets persisted even when changing panes
this should fix the sticky value bug. which changing values on double/float causes it to fix the value even when looking at other setting entries


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4601833610)

PR #1027, opened 2023-05-08
incorrect status bar fixes [CPP-927]

adrian-kong:
image

addresses RTK issue (conflicting variable in QML and js)

unsure if INS is fixed (might need replay file to verify)

### DIFF
--- a/console_backend/src/status_bar.rs
+++ b/console_backend/src/status_bar.rs
@@ -39,7 +39,7 @@ use crate::piksi_tools_constants::{
     ins_error_dict, ins_mode_dict, ins_type_dict, rtk_mode_dict, DR_MODE, EMPTY_STR, RTK_MODES,
 };
 use crate::shared_state::SharedState;
-use crate::types::{BaselineNED, GnssModes, PosLLH};
+use crate::types::{BaselineNED, Dops, GnssModes, PosLLH};
 use crate::utils::{bytes_to_kb, decisec_to_sec, serialize_capnproto_builder};
 
 enum AntennaStatus {
@@ -274,6 +274,7 @@ impl StatusBar {
                 .heartbeat_data
                 .lock()
                 .expect(HEARTBEAT_LOCK_MUTEX_FAILURE);
+            shared_data.ins_status_string = format!("0x{:<01x}", msg.flags);
             shared_data.ins_status_flags = ins_status_flags;
             shared_data.last_ins_status_receipt_time = last_ins_status_receipt_time;
         }
--- a/console_backend/src/status_bar.rs
+++ b/console_backend/src/status_bar.rs
@@ -39,7 +39,7 @@ use crate::piksi_tools_constants::{
     ins_error_dict, ins_mode_dict, ins_type_dict, rtk_mode_dict, DR_MODE, EMPTY_STR, RTK_MODES,
 };
 use crate::shared_state::SharedState;
-use crate::types::{BaselineNED, Dops, GnssModes, PosLLH};
+use crate::types::{BaselineNED, GnssModes, PosLLH};
 use crate::utils::{bytes_to_kb, decisec_to_sec, serialize_capnproto_builder};
 
 enum AntennaStatus {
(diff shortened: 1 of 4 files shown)


console_backend/src/tabs/solution_tab/solution_position_tab.rs
```diff
@@ -396,8 +398,6 @@ impl SolutionPositionTab {
         let dops_fields = msg.fields();
         self.table
             .insert(DOPS_FLAGS, format!("0x{:<03x}", dops_fields.flags));
-        self.table
-            .insert(INS_STATUS, format!("0x{:<01x}", self.ins_status_flags));
```
adrian-kong, line 400: not quite sure what this was doing in handle dops,
should INS status be updated based on DOPS updates?

console_backend/src/types.rs
```diff
@@ -1108,8 +1108,8 @@ impl GnssModes {
             GnssModes::NoFix => NO_FIX_LABEL,
             GnssModes::Spp => SPP,
             GnssModes::Dgnss => DGNSS,
-            GnssModes::Float => RTK,
-            GnssModes::Fixed => RTK,
+            GnssModes::Float => FLOAT,
+            GnssModes::Fixed => FIXED,
```
adrian-kong, line 1112: this addresses the same 'RTK' label in status bar issue

console_backend/src/status_bar.rs
```diff
@@ -274,6 +274,7 @@ impl StatusBar {
                 .heartbeat_data
                 .lock()
                 .expect(HEARTBEAT_LOCK_MUTEX_FAILURE);
+            shared_data.ins_status_string = format!("0x{:<01x}", msg.flags);
```
adrian-kong, line 277: basically how solution table inserts ins status string, did we want to replicate the logic from the solution table?

console_backend/src/types.rs
adrian-kong: moved to QML to fix issue

console_backend/src/status_bar.rs
```diff
@@ -39,7 +39,7 @@ use crate::piksi_tools_constants::{
     ins_error_dict, ins_mode_dict, ins_type_dict, rtk_mode_dict, DR_MODE, EMPTY_STR, RTK_MODES,
 };
 use crate::shared_state::SharedState;
-use crate::types::{BaselineNED, GnssModes, PosLLH};
+use crate::types::{BaselineNED, Dops, GnssModes, PosLLH};
```
adrian-kong, line 42:
```suggestion
use crate::types::{BaselineNED, GnssModes, PosLLH};
```

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4911083288)

silverjam:
Do we understand why this broke?

adrian-kong:
> Do we understand why this broke? Or did it ever work?
@silverjam 

think my PR https://github.com/swift-nav/swift-toolbox/pull/972 created this side effect.
variable naming conflicted with pyside and qml.
`set_rtk` in python class  and `rtk` in qml property which this changed the scope of variable assignments 😢 annoying bug thats super easy to overlook.
it might follow that we should consider naming qml properties uniquely to distinguish them from pyside variables in the future to prevent these types of weird behaviour

silverjam (approved):

console_backend/src/tabs/solution_tab/solution_position_tab.rs
silverjam: @john-michaelburke any ideas?

@adrian-kong can look at the old console? Maybe some hints there.  I'm not sure why we'd handle this in DOPS either.

john-michaelburke: Yeah purely an artifact of copy pasta:
```
https://github.com/swift-nav/piksi_tools/blob/7725f247b978124f3ee7d23d67a0995f3aaa6f24/piksi_tools/console/solution_view.py#L573
```

PR #1033, opened 2023-05-10
Add ntrip client [v4.0] [CPP-944]

adrian-kong:
- adds additional ntrip subtab to advanced tab
- adds`--enable-ntrip` flag, safe to merge since this feature is off by default
- static option (manual input) and dynamic (auto lat lon alt values) 
- output options, RTCM or SBP out
- bundles rtcm3tosbp binary under `/binaries/{OS}/rtcm3tosbp(.exe)`
- adds additional hidden until running status bar to capture amount of bytes

image

image


adrian-kong:
with regards to the options that we want to parse, is it something like...

general textfields that we must have:
URL, USERNAME, PASSWORD

* radio button, [GGA/CRA]
 
```
if GGA:
  add LAT, LON, ALT textfields
elif CRA:
  add REQ_COUNTER, AREA_ID, CORR_MASK, SOLN_ID text fields
```

i.e. main fields we want / capnproto struct is:
```
struct NtripConfig {
    url, username, password,
    (if gga then fill in...) lat, lon, alt
    (if cra then fill in...) req_counter, area_id, corr_mask, soln_id
}
```

add connect and disconnect button

on connect basically run main function from https://github.com/swift-nav/ntripping/blob/master/src/main.rs#L277 (maybe we should turn this into lib to reuse?)
on disconnect basically kills the thread ran in main.

- add some additional capnproto packets to send byte info, error log, ntrip config etc...

also havent touched ntrip before... didnt really understand what we wanted to do with the emitting output from ntripping.



@silverjam

console_backend/src/ntripping.rs
```diff
@@ -0,0 +1,247 @@
...
+                let lon_deg = lonn as u16;
+
+                let lat_min = (latn - (lat_deg as f64)) * 60.0;
+                let lon_min = (lonn - (lon_deg as f64)) * 60.0;
+
+                let lat_dir = if lat < 0.0 { 'S' } else { 'N' };
+                let lon_dir = if lon < 0.0 { 'W' } else { 'E' };
+
+                format!(
+                    "$GPGGA,{},{:02}{:010.7},{},{:03}{:010.7},{},4,12,1.3,{:.2},M,0.0,M,1.7,0078",
```
silverjam, line 129: This is an NMEA message that's generated and sent to the server, the `lat` and `lon` fields that are here will need to come from the SBP messages that the connected device sends us.

console_backend/src/ntripping.rs
```diff
@@ -0,0 +1,247 @@
...
+                    time, lat_deg, lat_min, lat_dir, lon_deg, lon_min, lon_dir, height
+                )
+            }
+            Message::Cra {
```
silverjam, line 133: We shouldn't need CRA for this feature

console_backend/src/ntripping.rs
```diff
@@ -0,0 +1,247 @@
...
+                    eprintln!("unpause error: {e}");
+                    return false;
+                }
+            }
+            true
+        }
+    })?;
+
+    transfer.borrow_mut().write_function(|data| {
+        if let Err(e) = std::io::stdout().write_all(data) {
```
silverjam, line 207: The data sent to `io::stdout()` here will need to be sent as input to the connected device.  For Mallard, this will be RTCM, for other devices that only support SBP input we'll either need to disable the NTRIP feature entirely or figure out how to bundle an RTCM to SBP converter.

silverjam:
> * radio button, [GGA/CRA]

Information for build GGA should come from two sources:
- Static, entered by the user
- Dynamic, from the lat/lon position emitted by the device

We don't need CRA for this feature.

silverjam:
> on connect basically run main function from https://github.com/swift-nav/ntripping/blob/master/src/main.rs#L277 (maybe we should turn this into lib to reuse?)

Could be useful eventually but I don't think we should pursue right now, let's see what's common between the two implementations before we make a decision to create a common library.

silverjam:
> also havent touched ntrip before... didnt really understand what we wanted to do with the emitting output from ntripping.

Ntrip is a protocol Skylark uses for delivering corrections to the device, the location of the device is sent to the server so that it can either: (1) fetch region specific error correction information; (2) model a virtual base station in order to provide error correction.  Data from Ntrip needs to be routed to the device in order for it to correct it's position.

resources/AdvancedTabComponents/NtripClientTab.qml
```diff
@@ -0,0 +1,147 @@
...
+                        let epoch = epochField.text;
+                        let lat = null;
+                        let lon = null;
+                        let alt = null;
+                        if (staticRadio.checked){
+                            lat = positionRepeater.itemAt(0).children[1].text;
+                            lon = positionRepeater.itemAt(1).children[1].text;
+                            alt = positionRepeater.itemAt(2).children[1].text;
+                        }
+                        backend_request_broker.ntrip_connect(url, username, password, epoch, lat, lon, alt);
```
adrian-kong, line 129: TODO: case where lat lon alt is empty string on static input.

src/main/resources/base/console_backend.capnp
```diff
@@ -452,6 +452,23 @@ struct AutoSurveyRequest {
...
+    lon @1: Float64;
+    alt @2: Float64;
+}
+
+struct NtripConnect {
+    url @0 :Text;
+    username @1 :Text;
+    password @2 :Text;
+    epoch @3 :UInt64;
+    position :union {
```
adrian-kong, line 466: maybe better / less space to take some out of bounds value as to be null rather constructing like this.

console_backend/src/connection.rs
```diff
@@ -167,6 +167,9 @@ fn conn_manager_thd(
                         &client_sender,
                     );
                     refresh_connection_frontend(&client_sender, &shared_state);
+                    {
+                        shared_state.lock().msg_sender = Some(msg_sender.clone());
```
adrian-kong, line 171: may be better to put this in connection.

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4997986861)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5009032648)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5041053122)

console_backend/src/cli_options.rs
```diff
@@ -199,6 +199,10 @@ pub struct CliOptions {
     /// Set the width of the main window.
     #[clap(long)]
     pub width: Option<u32>,
+
+    /// Disable ntrip client
+    #[clap(long)]
+    pub disable_ntrip: bool,
```
silverjam, line 205: This needs to be inverted to `--enable-ntrip` -- we want a way to force ntrip to be enabled so that we can test the client with other setups (such as chaining socat instances together).

silverjam: By default the ntrip client will be disabled unless we detect that we're connect to a PT device.

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5074379845)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5107216968)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5107965623)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5127899883)

adrian-kong:
https://github.com/swift-nav/swift-toolbox/releases/tag/v4.0.19-ntrip-test-6

working bundled rtcm3tosbp converter (tested on mac)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5266311196)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5268008566)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5273212792)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5284495431)

pcrumley (commented):
I'll leave a more formal review ASAP but I think we should store those binary files in LFS.

console_backend/src/utils.rs
```diff
@@ -16,6 +17,46 @@ use crate::errors::*;
...
+pub fn app_dir() -> crate::types::Result<PathBuf> {
+    std::env::current_exe()?
+        .parent()
+        .ok_or(anyhow::format_err!("no parent directory"))
+        .map(Path::to_path_buf)
+}
+
+/// Returns directory to packaged python, or workspace when ran locally in dev environment
+/// This is used to locate rtcm3tosbp
+pub fn pythonhome_dir() -> crate::types::Result<PathBuf> {
```
pcrumley, line 29: i definitely appreciate all the comments, which helps a lot but this feels a little hacky. Unfortunately I don't have a better solution off the top of my head. I would consider making `"py39"` a global const though, and try to link it to where-ever it is comes from when created the dev env folder

src/main/resources/base/console_backend.capnp
pcrumley: don't know enough bout capnp to comment

console_backend/src/ntrip_output.rs
```diff
@@ -0,0 +1,125 @@
...
+            cmd
+        } else if cfg!(target_os = "macos") {
+            let mut cmd = Command::new("sh");
+            let rtcm = pythonhome_dir()?
+                .join("binaries")
+                .join("mac")
+                .join("rtcm3tosbp")
+                .to_string_lossy()
+                .to_string();
+            warn!("running rtcm3tosbp from \"{}\"", rtcm);
```
pcrumley, line 72: info or debug level imo

pcrumley (commented):
I have just a few questions about this PR:
1) do we need to worry about gracefully shutting down and making sure things in the message channels are flushed to disk after the stop thing is sent? I tried to make in line where this could be a problem, but i realize now you'll probably just want to wait and then kill if it takes more than some set time to write.
2) when calling try_recv, that's not a blocking function, so this might spin pretty fast on some of the threads. It might be worth parking and unparking the threads. Might not be an issue if we expect the channels will always be full.

Other stuff was either clarifying questions or straightforward suggetsions.

console_backend/src/ntrip_tab.rs
```diff
@@ -0,0 +1,413 @@
...
+    after: u64,
+    epoch: Option<u64>,
+    crc: Option<u8>,
+    #[serde(flatten)]
+    message: Message,
+}
+
+fn default_after() -> u64 {
+    10
+}
```
ziwift, line 131: better to define this as a const val, or at least `const fn` if you really want to use fn

console_backend/src/ntrip_output.rs
```diff
@@ -0,0 +1,125 @@
...
+        let running = self.running.clone();
+        thread::spawn(move || loop {
+            if !running.get() {
+                break;
+            }
+            if let Ok(data) = in_rx.try_recv() {
+                if let Err(e) = out.write(&data) {
+                    error!("failed to write to device {e}");
+                }
+            }
```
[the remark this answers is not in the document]
adrian-kong: hmm yeah good idea, will just let this gracefully terminate via channel drop

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5306756529)

console_backend/src/utils.rs
adrian-kong: yeah this is really hacky, think the better way might be passing things via maketoml
but this should be ok since swift console executable must be within these folders.
*also didn't want to spend a bunch of time tinkering the maketoml, iirc a bunch of hardcoded py39 paths there when trying to upgrade to py311...

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5306953790)

silverjam (approved):

pcrumley (approved):
LGTM... however do you need to port the unrelated bugfixes in resources/SettingsTabComponents/SettingsPane.qml file in the other PR to this one? or were those bugs introduced after v4.0

silverjam:
Agree with @pcrumley -- let's make sure we don't miss anything from https://github.com/swift-nav/swift-toolbox/pull/1088/files#diff-426f1e106c29afd7217275dd47a806603df0770a48d007397e430ee887fc1949 (if these are needed on this branch).

adrian-kong:
the changes for `parents` -> `parent` seem to be a typo causing `ReferenceError` (introduced later)
the changes for height seems to be a recurring issue i.e. the settings pane cutting off or scroll bar doesn't adjust to the amount of content on page (will revert this unrelated changes (doesn't seem to fix the issue) for both branches and logged a ticket: https://swift-nav.atlassian.net/browse/CPP-953 if this ever becomes an issue again)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5383510731)

adrian-kong:
moved to SCD board https://swift-nav.atlassian.net/browse/SCD-3

PR #1046, opened 2023-05-17
use epoch

adrian-kong:


console_backend/src/ntrip_tab.rs
```diff
@@ -122,7 +121,7 @@ impl Command {
 impl std::fmt::Display for Command {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let now = self.epoch.map_or_else(SystemTime::now, |e| {
-            SystemTime::UNIX_EPOCH + Duration::from_secs(e.into())
+            SystemTime::UNIX_EPOCH + Duration::from_secs(e)
```
adrian-kong, line 124: just wanted to verify if this is correct w.r.t tow conversion from PosLLH @silverjam

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/4998283182)

console_backend/src/ntrip_tab.rs
silverjam: Maybe, what is the behavior change that's happening here?

silverjam: I think in general we can use the time from the device or from the system, but using the time from the device is more likely to be correct

adrian-kong: > Maybe, what is the behavior change that's happening here?

to confirm this calculation was correct:
```rust
// fields.tow corresponds to MsgPosLLH#tow
        let tow_s = fields.tow / 1000.0;
// taken from MsgPosLLH#gps_time(), wn = 0, just substituting tow 
        guard.epoch = match sbp::time::GpsTime::new(0, tow_s) {
// and we want the utc seconds to add UNIX_EPOCH
            Ok(gps_time) => Some(gps_time.to_utc_hardcoded().seconds().round() as u64),
            Err(_) => None,
        };
```

silverjam: I think for it to be correct, we'll need to source `wn` from somewhere

silverjam: @ljbade if we want to convert from gps time to utc, we need to supply the week number, correct?

ljbade: hmm yeah the number of leap seconds between GPS and UTC depends on the date, basically if you have some date before the last leap second was implemented it will have a different offset to after
I recall somewhere in our code we have a list of these dates

ljbade: the MsgPosLLH should have a wn field

adrian-kong: ah it seems:
```
MSG_POS_LLH — 0x020A — 522
...
The full GPS time is given by the preceding MSG_GPS_TIME with the matching time-of-week (tow).
```

adrian-kong: updated to take in week number from GpsTime (might not be ideal since it depends on two packets to construct the correct epoch)

silverjam (approved):

PR #1054, opened 2023-05-22
use combo boxes [CPP-944]

adrian-kong:
image

https://support.swiftnav.com/support/solutions/articles/44001850766-receiver-configuration-for-skylark-operation

Region | Bands L1/L2 | Bands L1/L5
-- | -- | --
North America | na.l1l2.skylark.swiftnav.com | na.l1l5.skylark.swiftnav.com
Europe | eu.l1l2.skylark.swiftnav.com | eu.l1l5.skylark.swiftnav.com
Asia Pacific | ap.l1l2.skylark.swiftnav.com | ap.l1l5.skylark.swiftnav.com

Port (2101 / 2102)
Mountpoint (OSR/MSM5)

editable combo boxes

### DIFF
--- a/resources/AdvancedTabComponents/NtripClientTab.qml
+++ b/resources/AdvancedTabComponents/NtripClientTab.qml
@@ -40,21 +40,21 @@ Item {
                     Layout.fillWidth: true
                     editable: true
                     model: ListModel {
-                        ListElement { text: "OSR" }
-                        ListElement { text: "MSM5" }
-                    }
-                    validator: stringValidator
-                    onAccepted: mountpoint = currentText
-                }
-                ComboBox {
-                    Layout.fillWidth: true
-                    editable: true
-                    model: ListModel {
                         ListElement { text: "2101" }
                         ListElement { text: "2102" }
                     }
                     validator: intValidator
                     onAccepted: port = currentText
+                }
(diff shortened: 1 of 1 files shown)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5042517743)

resources/AdvancedTabComponents/NtripClientTab.qml
```diff
@@ -13,15 +13,94 @@ Item {
...
+            RowLayout {
+                ComboBox {
+                    Layout.fillWidth: true
+                    editable: true
+                    validator: stringValidator
+                    onAccepted: url = currentText
+
+                    model: ListModel {
+                        ListElement {
+                            text: "na.l1l2.skylark.swiftnav.com"
```
silverjam, line 33: This looks great, but I don't think it's what we need for this feature, we want to allow users to connect other NTRIP providers, and we (internally) want to be able to connect to internal staging or preproduction environments.

If we had a way to only conditionally show this combo box UI then it could be great.  For instance of someone selected "Skylark" as the ntrip type vs "Other".

For now I would just leave this as an experiment.

PR #1065, opened 2023-05-29
better status bar

adrian-kong:
https://github.com/swift-nav/swift-toolbox/assets/35755741/dc62e006-4a3b-448c-8f6f-f000f79b53d7

changed to per second

### DIFF
--- a/console_backend/src/ntrip_tab.rs
+++ b/console_backend/src/ntrip_tab.rs
@@ -184,6 +184,30 @@ fn get_commands(
     Ok(Box::new(iter::once(first).chain(rest)))
 }
 
+#[derive(Default)]
+struct Progress {
+    ul_tot: f64,
+    dl_tot: f64,
+    ul_tot_old: f64,
+    dl_tot_old: f64,
+}
+
+impl Progress {
+    /// Fetch changed download and modify last changed
+    pub fn tick_dl(&mut self) -> f64 {
+        let diff = self.dl_tot - self.dl_tot_old;
+        self.dl_tot_old = self.dl_tot;
+        diff
+    }
+
+    /// Fetch changed upload and modify last changed
+    pub fn tick_ul(&mut self) -> f64 {
+        let diff = self.ul_tot - self.ul_tot_old;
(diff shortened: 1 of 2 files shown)


swiftnav_console/main.py
```diff
@@ -478,9 +478,16 @@ def _process_message_buffer(self, buffer):
 
             up = m.statusBarStatus.ntripUpload
             down = m.statusBarStatus.ntripDownload
+            down_units = "B/s"
+
+            if down >= 1000:
+                down /= 1000
+                down = round(down, 1)
+                down_units = "KB/s"
+
```
adrian-kong, line 487: upload bytes is roughly 80bytes, download is the only value that needs the conversion sometimes

silverjam (approved):

console_backend/src/ntrip_tab.rs
```diff
@@ -219,18 +243,37 @@ fn main(
...
+            heartbeat.set_ntrip_dl(progress.tick_dl());
+        }
+        thread::park_timeout(Duration::from_secs(1))
+    });
+
     transfer.borrow_mut().progress_function({
         let rx = &rx;
         let transfer = Rc::clone(&transfer);
-        move |_dltot, _dlnow, _ultot, _ulnow| {
+        move |_dlnow, dltot, _ulnow, ultot| {
```
adrian-kong, line 266: naming are different https://docs.rs/curl/latest/src/curl/easy/handle.rs.html#359
```
    /// The callback gets told how much data libcurl will transfer and has
    /// transferred, in number of bytes. The first argument is the total number
    /// of bytes libcurl expects to download in this transfer. The second
    /// argument is the number of bytes downloaded so far. The third argument is
    /// the total number of bytes libcurl expects to upload in this transfer.
    /// The fourth argument is the number of bytes uploaded so far.
```
doesn't seem like first/third argument works tho i.e. only gives 0

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5107471394)

PR #1066, opened 2023-05-29
bundle rtcm converter [CPP-949]

adrian-kong:
bundling RTCM converter and piping data to the process, returning the output to msg_sender

will also need to notarize binaries and include them in the packaging process... but for now trying to figure out why sending data to device does nothing


console_backend/src/ntrip_tab.rs
```diff
@@ -284,8 +285,11 @@ fn main(
         }
     })?;
 
-    transfer.borrow_mut().write_function(|data| {
-        if let Err(e) = msg_sender.write_all(data) {
+    let (rtcm_tx, rtcm_rx) = channel::unbounded::<Vec<u8>>();
+    let rtcm2sbp = RtcmConverter::new(rtcm_rx);
+    rtcm2sbp.start(msg_sender);
+    transfer.borrow_mut().write_function(move |data| {
+        if let Err(e) = rtcm_tx.send(data.to_owned()) {
```
adrian-kong, line 292: send rtcm3tosbp output back directly to msg_sender

console_backend/src/rtcm_converter.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+impl RtcmConverter {
+    pub fn new(in_rx: Receiver<Vec<u8>>) -> Self {
+        Self { in_rx }
+    }
+
+    pub fn start<W: Write + Send + 'static>(&self, mut out: W) {
+        println!("starting rtcm3tosbp");
+        let mut child = Command::new("./binaries/mac/rtcm3tosbp")
```
adrian-kong, line 18: TODO: will need to clear this up later but for testing purposes right now...

console_backend/src/ntrip_tab.rs
adrian-kong: still doesn't seem to be doing anything... maybe `MsgSender` is incorrect place to send?

console_backend/src/ntrip_tab.rs
```diff
@@ -247,11 +249,6 @@ fn main(
     let progress_clone = progress.clone();
     let running_clone = is_running.clone();
```
adrian-kong, line 250:
```suggestion
```

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5129087821)

resources/AdvancedTabComponents/NtripClientTab.qml
```diff
@@ -125,6 +125,25 @@ Item {
                 enabled: !connected
             }
 
+
```
adrian-kong, line 128:
```suggestion
```

console_backend/src/ntrip_tab.rs
adrian-kong: resolved, this seems to be a device issue. works fine sending SBP back.

console_backend/src/types.rs
```diff
@@ -72,8 +75,13 @@ impl MsgSender {
     const LOCK_FAILURE: &'static str = "failed to aquire sender lock";
 
     pub fn new<W: Write + Send + 'static>(writer: W) -> Self {
+        let file = File::create("ntrip_output").ok();
+        if file.is_none() {
+            error!("could not create file!");
+        }
```
adrian-kong, line 81: This is for the purpose of debugging at the moment it seems
- Sending RTCM not working for Piksi Triple device.
- Works fine with strsvr just not via ntrip tab.
- Sending RTCM converted to SBP data works on other devices.

resources/console_resources.qrc
```diff
@@ -137,5 +137,8 @@
     <file>MainTab.qml</file>
     <file>TabInfoBar.qml</file>
     <file>PositionLoopAnimation.qml</file>
+    <file>binaries/linux/rtcm3tosbp</file>
+    <file>binaries/mac/rtcm3tosbp</file>
+    <file>binaries/win/rtcm3tosbp.exe</file>
```
adrian-kong, line 142: Not sure where binaries actually go - need to look more into how to actually distribute this🤔

resources/AdvancedTabComponents/NtripClientTab.qml
```diff
@@ -125,6 +125,24 @@ Item {
...
+                        text: "RTCM"
+                    }
+
+                    ListElement {
+                        text: "SBP"
+                    }
+
+                }
+
+            }
```
adrian-kong, line 144: Might be better to do checkbox instead since we only allow either RTCM or SBP. Not sure if any future elements will be added.

src/main/resources/base/console_backend.capnp
```diff
@@ -470,6 +470,7 @@ struct NtripConnect {
         pos @4 :Position;
         none @5 :Void;
     }
+    outputType @6: Text;
```
adrian-kong, line 473: Considering enum or just boolean `use_sbp` to save some negligible amount of parsing and capnproto allocations 🔨

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5141075613)

console_backend/src/ntrip_output.rs
```diff
@@ -0,0 +1,103 @@
...
+            cmd
+        } else if cfg!(target_os = "macos") {
+            let mut cmd = Command::new("sh");
+            cmd.args(["-c", "resources/binaries/mac/rtcm3tosbp"]);
+            cmd
+        } else {
+            let mut cmd = Command::new("sh");
+            cmd.args(["-c", "resources/binaries/win/rtcm3tosbp"]);
+            cmd
+        };
```
adrian-kong, line 64: MacOS is tested working.

(Also thinking alternative to embedding rtcm3tosbp by allowing selectable binary that consumes RTCM stdin and outputs stdout - then provide the binaries auto-download and just let users select path. If there are any changes to rtcm3tosbp or alternative format to send to device)

console_backend/src/ntrip_output.rs
```diff
@@ -0,0 +1,103 @@
...
+        let mut child = child
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
```
adrian-kong, line 68: Does not handle rtcm3tosbp errors. Not sure if necessary.

console_backend/src/ntrip_tab.rs
```diff
@@ -245,27 +258,29 @@ fn main(
...
+        move || loop {
+            {
+                if !running.get() {
+                    return false;
+                }
             }
+            {
+                let mut progress = progress_clone.lock().unwrap();
+                heartbeat.set_ntrip_ul(progress.tick_ul());
+                heartbeat.set_ntrip_dl(progress.tick_dl());
```
adrian-kong, line 272: This only shows NTRIP client data (sent and received in RTCM).
Does not show converted RTCM to SBP size that is being sent to device.

console_backend/src/ntrip_tab.rs
```diff
@@ -345,31 +360,40 @@ impl NtripState {
...
+                if let Err(e) = output_converter.start(msg_sender).and(main(
+                    heartbeat.clone(),
+                    options,
+                    last_data,
+                    running.clone(),
+                    conv_tx,
+                )) {
+                    error!("{e}");
+                }
+                println!("terminating...");
```
adrian-kong, line 380:
```suggestion
```

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5141224860)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5150525335)

console_backend/src/ntrip_output.rs
silverjam: We'll need to empty these pipes to some where, otherwise it will eventually fill up  and the process will block (and appear to lock up) when it becomes full.

So anything that we're not explicitly writing or reading to either needs to be `Stdio::null()` or we need to add code to empty the pipe regularly.

console_backend/src/types.rs
adrian-kong: File can't be created in `.app` 🤔

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5172329152)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5182873152)

console_backend/src/ntrip_output.rs
```diff
@@ -0,0 +1,109 @@
...
+        });
+        Ok(())
+    }
+
+    /// Runs rtcm3tosbp converter
+    fn output_sbp<W: Write + Send + 'static>(&mut self, mut out: W) -> Result<()> {
+        let curr_dir = std::env::current_dir()?;
+        let curr_dir = curr_dir.to_str().expect("current directory not found");
+        let cmd = format!("ls {curr_dir}");
+        println!("{}", cmd);
```
adrian-kong, line 55:
```
ls /

Applications
Library
System
Users
Volumes
bin
cores
dev
etc
home
opt
private
sbin
tmp
usr
var
```
not relative to `.app`, 

this pr uses selector instead of bundling
https://github.com/swift-nav/swift-toolbox/pull/1076

silverjam:
ping @swift-nav/devinfra

console_backend/src/ntrip_output.rs
adrian-kong: hmm am stuck on whether there is an easy way to locate the path for bundled files i.e. `Swift Console.app\Contents\...` on mac

alternatively really simple, [here is a test released version](https://github.com/swift-nav/swift-toolbox/pull/1076) with user provided binary (to save the packaging trouble)

or maybe like - `include_bytes`, when ran write the bytes to file to execute.

or maybe pass path from python? https://github.com/swift-nav/swift-toolbox/blob/3f4874edafdae9a55073065bdf13885e536c953e/swiftnav_console/main.py#L584

any ideas? 🤔 
@silverjam @swift-nav/devinfra

silverjam: https://doc.rust-lang.org/std/env/fn.current_exe.html

This returns a PathBuf, wouldn't that be sufficient to locate a helper binary?

adrian-kong: > https://doc.rust-lang.org/std/env/fn.current_exe.html
> 
> This returns a PathBuf, wouldn't that be sufficient to locate a helper binary?

yeah that returns
```
"/Users/adriankong/IdeaProjects/v/swift-toolbox/py39/bin/python3"
```
path to python on build oddly enough

silverjam: We've already solved this problem for other things (like windowpos exe and qt web engine process) we should be able to adapt one of those existing solutions.

silverjam: > not relative to `Swift Console.app` when ran in release on mac

The current working directory of the app is going to vary widely depending on how it's launched, so this isn't a viable solution.

adrian-kong: > It's something that's going to need to be handled differently if we're running a development build versus a packaged build.
> 
> Similar problems are handled by our `entrypoint` code:
> 
> https://github.com/swift-nav/swift-toolbox/blob/3f4874edafdae9a55073065bdf13885e536c953e/entrypoint/src/main.rs#L66

I'll give this one a try 👍

silverjam:
I was finally able to get this to build locally, not totally sure what was going on but it looks like somewhere along the lines the swiftnav Rust library stopped building with the version of clang that I'm able to install locally.  The master version of swiftnav-rs has an updated version of bindgen that works, so I pushed a new release of swiftnav-rs and was finally able to build locally.

Thus, I was able to get this running, connect to NTrip and saw data flowing -- the app stayed connected to NTrip too.  All that was good, but I was not able to see any remote observations, so things are not completely happy yet.  My guess is there is an issue either locating or running `rtcm3tosbp.exe` which is causing the RTCM data to not be converted correctly.

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5247916350)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5253595146)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5261404047)

adrian-kong:
will merge this into ntrip_client branch

PR #1076, opened 2023-06-07
use path provided rtcm converter for ntrip output [CPP-949]

adrian-kong:
having issues bundling rtcm3tosbp into qt app (unsure location of the binary),
this allows user to select any binary that consumes RTCM in stdin and grabs stdout when provided.

also just to allow testing purposes

image

with no converter provided, will just send RTCM data


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5195529270)

console_backend/src/ntrip_output.rs
```diff
@@ -1,32 +1,46 @@
...
     pub fn start<W: Write + Send + 'static>(&mut self, out: W) -> Result<()> {
         self.running.set(true);
-        match self.output_type {
-            OutputType::RTCM => self.output_rtcm(out),
-            OutputType::SBP => self.output_sbp(out),
+        if self.binary_path.is_empty() {
+            return self.output_rtcm(out);
+        }
+        let path = self.binary_path.clone();
+        let meta = fs::metadata(path.clone());
```
pcrumley, line 30: unnecessary clone
```suggestion
        let meta = fs::metadata(&path);
```

console_backend/src/ntrip_output.rs
```diff
@@ -1,32 +1,46 @@
...
 use std::io::Write;
 use std::process::{Command, Stdio};
-use std::{io, thread};
+use std::{fs, io, thread};
 
 pub struct MessageConverter {
     in_rx: Receiver<Vec<u8>>,
     running: ArcBool,
-    output_type: OutputType,
+    binary_path: String,
```
pcrumley, line 12: This is totally fine, but i do feel like it might be a good usecase for UTF8Path & the camino crate: https://docs.rs/camino/latest/camino/struct.Utf8Path.html

console_backend/src/ntrip_tab.rs
```diff
@@ -352,6 +336,7 @@ impl NtripState {
         msg_sender: MsgSender,
         mut heartbeat: Heartbeat,
         options: NtripOptions,
+        binary: String,
```
pcrumley, line 339: to me this function arg is not the best name. And it doesn't make since for binary to be String, i'd expect it to be Vec<u8>

pcrumley: and should it be an Option

adrian-kong: oops binary is probably not the best name,
maybe `converter_path` would be better

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5253657520)

silverjam:
Close?  I don't think we need this anymore.

PR #1115, opened 2023-07-22
Support PGM in tracking signals tab[DIP-64]

john-michaelburke:
# Implements
* Adds handle_msg_status_report to the TrackingSignalsTab to discern whether the source device is a PGM.
* If the device is a PGM, we disable the processing of MsgTrackingState and MsgMeasurementState in favor of MsgObs.

# Concerns
* There is a slight visual artifact at the beginning of the stream which is due to the delayed reception of the MsgStatusReport but after the message is received things look normal.

[Screencast from 07-21-2023 07:10:02 PM.webm](https://github.com/swift-nav/swift-toolbox/assets/43353147/1afff5f3-e7b5-4940-a22e-491d62b54b4d)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5632880649)

silverjam (approved):

adrian-kong (approved):

PR #1116, opened 2023-07-22
Handle bad log messages, add log-level,log-stdout cli args

john-michaelburke:
## Implements
* Handle bad log messages. When the console starts with Debug logging enabled there were a few messages that have escaped quotation chars which would cause the console to crash so this regex should remove those.
* Add cli argument for setting the log level, on master the default log level was WARNING so no lower-level warnings would get dumped to a file/console until you start the console and then set the level. I think adding this as an argument is pretty reasonable, but if there is pushback, we can set it as hidden.
* Add cli argument for dumping logs to stdout. I find it is much easier to debug an issue if we can dump the logs to a terminal window instead of a separate log file every time the console starts. Again if there is pushback we can set the argument to hidden.

### DIFF
--- a/console_backend/src/cli_options.rs
+++ b/console_backend/src/cli_options.rs
@@ -85,6 +85,14 @@ pub struct CliOptions {
     #[clap(long)]
     pub log_console: bool,
 
+    /// Log to stdout.
+    #[clap(long)]
+    pub log_stdout: bool,
+
+    /// Set log level.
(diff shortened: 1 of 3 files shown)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/5632881521)

silverjam (approved):

PR #1186, opened 2023-09-13
fix taplo version@0.8.1 & resolve compilation errors with --locked

pcrumley:
This PR installs taplo-cli with the tip described [here](https://taplo.tamasfe.dev/cli/installation/cargo.html). Namely you should use `--locked` if you are getting compliation errors (and we were).

I also chose to lock taplo to v0.8.1. It would be trivial to switch this to take the latest.


adrian-kong (approved):
looks like its working

PR #1227, opened 2023-11-02
disable refresh on record

adrian-kong:
Not too sure if this breaks things, disables the force refresh on recording button.


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/6742517048)

PR #1256, opened 2024-07-24
Collection of fixes to get CI running.

john-michaelburke:
* Clippy fixes.
* Fixed an issue with new cargo make duckscript. We were character escaping a boolean which seems to be invalid now. Fixing this now allows the correct collection of python files to be globbed for python checks to run.
* Uprev macos-11 to macos-12 due to deprecation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,7 @@ jobs:
     name: Backend Benchmarks
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/10085067359)

PR #1258, opened 2024-07-31
Fix map view feature

john-michaelburke:
## Existing Bug
* Currently if the user attempts to use `--enable-map` on the master branch it crashes the application.
image

## Implements
* Re-enables the map feature such that the tab is always visible.
* Removes the "enableMap" feature, as it is not working in the existing implementation.
* To allow developers the ability to see the map without the token being encoded, I add a retry mechanism where it will attempt to load whatever the user had locally stored in `MAPBOX_TOKEN` environment variable. 
* Mutes a few verbose logs relevant to the webengine loading.
* [mapbox allows up to 50k map loads for free](https://www.mapbox.com/pricing#map-loads-for-web) per month so it seems relatively harmless to always include the feature.

## Options
* I don't think the conditionally loading with the Globals.enableMap feature is the best solution, and would require a bit of QML finesse to get it working properly (mainly hiding the tab when not enabled). So one option which this PR enables is to keep it on all the time.
* Second option is removing it entirely.
* Third option would be to try and rework the QML to get conditional loading working which may have a bit of a learning curve. Lots of trial and error.

## Testing
* I built a test tag `v4.1.21-jm-test` and here is the map feature in action.

https://github.com/user-attachments/assets/2fb801e3-2a89-4b42-a718-ad91082a5650

### DIFF
--- a/console_backend/src/cli_options.rs
+++ b/console_backend/src/cli_options.rs
@@ -109,10 +109,6 @@ pub struct CliOptions {
     #[clap(long)]
     pub show_file_connection: bool,
 
-    /// Enable map.
-    #[clap(long)]
-    pub enable_map: bool,
-
     /// Enable ntrip client
     #[clap(long)]
     pub enable_ntrip: bool,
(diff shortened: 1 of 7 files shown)


swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/10173410189)

swiftnav-travis:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/10186972916)

PR #1262, opened 2026-06-03
Fix glonass measurement state

jackleckert:
This PR enables to fix the tracking of GLONASS measurements state. Instead of using indexes to associate PRNs with corresponding frequency channel number, it now relies on `state.mesid.sat` slot number.

### DIFF
--- a/console_backend/src/tabs/tracking_tab/tracking_signals_tab.rs
+++ b/console_backend/src/tabs/tracking_tab/tracking_signals_tab.rs
@@ -575,13 +575,38 @@ mod tests {
         );
         assert_eq!(tracking_signals_tab.glo_slot_dict.len(), 1);
         assert_eq!(
-            tracking_signals_tab.glo_slot_dict[&0_i16],
+            tracking_signals_tab.glo_slot_dict[&(glo_sat_under_one_hundred as i16)],
             glo_sat_under_one_hundred as i16
         );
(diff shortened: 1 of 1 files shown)


pcrumley (changes requested):
I made the rust code more idiomatic & hopefully pass the linter but I also think this is something which really should have at lease one unit test because its kind of confusing what the logic is doing.

pcrumley:
@jackleckert this is failing a unit test but i don't think it is from my refactor, if you look at your PR you'll see it was broken before it even got to this step.

swiftnav-svc-devinfra:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/29521325012)

pcrumley (approved):
I have no idea why this change was made or the context of the change. However I am approving because my concern about the unit tests were fixed and change seems fine enough. I'd encourage you to get a second opinion from someone who knows more about this.

jackleckert:
> I have no idea why this change was made or the context of the change. However I am approving because my concern about the unit tests were fixed and change seems fine enough. I'd encourage you to get a second opinion from someone who knows more about this.

Thanks @pcrumley! @ebethon I have tagged you as well to review the GLO measurement state change itself.

ebethon (approved):

PR #1264, opened 2026-07-15
add dependabot tracking of github actions

pcrumley:


jackleckert (commented):
Do we want to enable automatic PR merging? seems like we have a handful of dependabot PRs opened since 2023.

pcrumley:
> Do we want to enable automatic PR merging? seems like we have a handful of dependabot PRs opened since 2023.

i think that is a good idea but better to add as a separate pr

jackleckert:
Sounds good! Don't we want to apply group updates to github actions as well? I wonder if all releases are stable as well...

PR #1267, opened 2026-07-21
CPP-967: always show File connection input, remove dead --show-file-connection flag

jackleckert:
## Summary
- Always show the **File** connection option in the console; drop the `--show-file-connection` gate
- The Browse (`...`) button and `*.sbp` picker already existed — they now appear with the File tab.
- Remove the now-dead `showFileConnection` global and the `--show-file-connection` flag (Python argparse + Rust clap).

## Test plan
- `cargo check -p console-backend` green; `qmllint` clean; Python `py_compile` OK.
- Regenerated the QML resource bundle (`makers generate-resources`).
- Manual (please confirm on your machine): launch without the flag → File radio visible → pick a `.sbp` via Browse → Connect.

## New behavior
Screenshot 2026-07-21 at 12 26 11 PM

## Notes
- No behavior change to the file-connection backend path (`connect_file`).
- Building `console-backend` here needed `CMAKE_POLICY_VERSION_MINIMUM=3.5` (pre-existing cmake-4 env issue, unrelated to this change).

PR #1294, opened 2026-08-21
Replace broken Mapbox map with Leaflet + Stadia Maps

badaimarcher:
## Summary
- Replaces the console's Mapbox GL JS map (broken on `main`, gated behind a fragile `--enable-map` flag) with Leaflet + Stadia Maps tiles, matching the approach already used by the internal Partner Portal.
- Removes `--enable-map` entirely, mirroring how `--show-file-connection` was retired in #1267 — the Map tab is now always shown. `--enable-map` remains in the Python arg parser only to emit a deprecation error if passed.
- Fixes a lon/lat ordering bug in the position-update handler (verified against the capnp schema and Rust serialization).
- Drops a leftover Mapbox-era token-obfuscation helper that silently corrupted any plain API key fed through it.
- Documents the new `STADIA_TOKEN` env var in the README, including a pointer to Stadia's free tier for local dev.

## Test plan
- [x] `cargo check -p console-backend` compiles cleanly after the `cli_options.rs` field removal
- [x] `cargo make generate-resources` succeeds and correctly substitutes `STADIA_TOKEN`
- [x] `python -m py_compile swiftnav_console/main.py` passes
- [x] `qmllint` shows no new warnings on changed QML files beyond pre-existing environment-limitation noise present elsewhere in the codebase
- [x] Verified via `cargo make run` with a personal Stadia Maps key: tiles render, map JS receives position updates over `QWebChannel` correctly
- [ ] Full manual pass (breadcrumb trail, per-mode ellipses, protection-level ellipse, focus-toggle pan, clean clear on disconnect/reconnect) — tiles confirmed rendering; recommend a final pass with a live/replayed solution stream before merge

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


swiftnav-svc-devinfra:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/32521681549)

swiftnav-svc-devinfra:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/32771219574)

swiftnav-svc-devinfra:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/32774991522)

jackleckert:
@badaimarcher could we keep Mapboxgl + Stadia maps?
This combination is what is currently implemented for the partner portal (curlew repo). The mapbox GL uses WebGL to render the map, which gives a much smoother interactive map compared to leaflet.

resources/SolutionTabComponents/SolutionMapTab.qml
```diff
@@ -15,7 +15,7 @@ Item {
     SolutionMap {
         id: solutionMap
         WebChannel.id: "currPos"
-        signal recvPos(int id, double lat, double lng, double hAcc)
+        signal recvPos(int id, double lon, double lat, double hAcc)
```
jackleckert, line 18: why change the order?

resources/web/map/js/trajectory_raw.js
```diff
@@ -202,40 +81,32 @@ new QWebChannel(qt.webChannelTransport, (channel) => {
...
+            fillOpacity: 0.3,
+        }).addTo(routeGroups[id]);
+        crumbLine.addLatLng(pos);
+
+        if (!currentMarker) currentMarker = L.marker(pos).addTo(map);
+        else currentMarker.setLatLng(pos);
+
         if (!startMarker) {
-            startMarker = new mapboxgl.Marker().setLngLat(pos).addTo(map);
+            startMarker = L.marker(pos).addTo(map);
```
jackleckert, line 98: could we use different colors or shapes for the start and current Maker? Right now it is hard to distinguish both.

jackleckert (commented):
Looks mostly good, just a couple comments:
- I tried running the leaflet map and this confirmed the experience I had previously with other tools: the map zooming in/out is not smooth and renders by discrete levels. Also any overlay is "jumping" around before getting to the right place again. This issue was only removed when I switched to using another library like `mapbox gl`, or its free open-source version `maplibre gl`.
- The new `STADIA_TOKEN` will need to be added to our private repo there in a separate PR: https://github.com/swift-nav/swift-toolbox-ci/tree/main/obfuscate-token

PR #1296, opened 2026-09-17
Document swift console platform support in README

jackleckert:


dgburr:
> Build to support, but not verified, on Apple Silicon.

Shouldn't that be "Built to support, but not verified, on Apple Silicon"?  And why can't we get someone to test it on Apple Silicon?  Surely there is at least one person at Swift with a relatively recent macbook?

PR #1297, opened 2026-09-17
Jesse/maplibre map prototype

badaimarcher:
Map enabled console prototype using maplibre-GL and stadia vector style tiles


swiftnav-svc-devinfra:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/35268374270)

jackleckert:
@badaimarcher I personally largely prefer this one 👍🏼

swiftnav-svc-devinfra:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/35655877311)

resources/web/map/js/trajectory_raw.js
```diff
@@ -202,15 +214,15 @@ new QWebChannel(qt.webChannelTransport, (channel) => {
         }
     });
 
-    chn.recvPos.connect((id, lat, lng, hAcc) => {
-        const pos = [lat, lng], rX = hAcc / 1000;
+    chn.recvPos.connect((id, lon, lat, hAcc) => {
```
jackleckert: are `lat` `lon` expected to switch to `lon` `lat`?

jackleckert (approved):
LGTM. The stadia token will also need to be added to the private [swift-toolbox-ci](https://github.com/swift-nav/swift-toolbox-ci/tree/main/obfuscate-token) repo in a separate PR.

swiftnav-svc-devinfra:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/35667272719)

resources/web/map/js/trajectory_raw.js
badaimarcher: From claude:

`The parameter names changed, but the actual values passed in were never swapped — the old code had them mislabeled, and I corrected the labels
  to match reality.

  Here's the evidence chain I traced back when doing the original Leaflet port:

  - console_backend.capnp defines Point{x, y}. In solution_position_tab.rs:779, the tuple pushed is (lon_values[jdx], lat_values[jdx]), and later
  that tuple is destructured as (x, y) and written with set_x(*x); set_y(*y). So on the wire, Point.x = lon, Point.y = lat.
  - swiftnav_console/solution_map.py's send_pos emits recvPos.emit(idx, pos.x, pos.y, status.hAcc) — meaning the true argument order is (idx, lon,
  lat, hAcc).

  The old Mapbox-era JS wrote chn.recvPos.connect((id, lat, lng, hAcc) => { const pos = [lat, lng], ... }). Given the true order above, lat (2nd
  param) was actually receiving the longitude value, and lng (3rd param) was actually receiving the latitude value — the names were backwards. It
  happened to still work for Mapbox, because pos = [lat, lng] just forwarded the arguments in the order they arrived — [arg1, arg2] = [lon, lat]
  in truth — which coincidentally matches GeoJSON's [lng, lat] convention that Mapbox/MapLibre wants. Right output, wrong variable names.

  I renamed the params to (id, lon, lat, hAcc) to reflect what they actually are. This wasn't just cosmetic — it mattered when porting to Leaflet,
  which wants [lat, lng] (the opposite order): if I'd trusted the old mislabeled names literally instead of tracing the real values, I'd have
  built the position array wrong and every marker would've landed at swapped coordinates. For MapLibre (this branch), since it uses the same [lng,
  lat] GeoJSON convention as the original Mapbox code, the fix meant pos = [lon, lat] — same numeric result as before, just now the code says
  what it means instead of lying about it.`

PR #1298, opened 2026-09-18
CPP-969: show "Swift Console" and version in window title before conn…

badaimarcher:
…ecting

The title bar was bound solely to statusBar.title, which stays empty until a device connection sends heartbeat data. Fall back to "Swift Console <version>" so the title is populated from startup.


jackleckert (approved):

swiftnav-svc-devinfra:
Frontend and Release Workflow Started [here](https://github.com/swift-nav/toolbox-backend-ci/actions/runs/35667220200)

